PR #12, opened 2021-11-22
Remove KVault dependency that contains androidx.

AfanasievAnton:


transport/src/iosMain/kotlin/com/genesys/cloud/messenger/transport/util/TokenStoreImpl.kt
```diff
@@ -1,11 +1,10 @@
...
-import com.liftric.kvault.KVault
 import platform.Foundation.NSUUID
 
 internal class TokenStoreImpl(storeKey: String) : TokenStore {
 
-    private val store = KVault(storeKey)
+    private val store = Vault(storeKey)
     override val token: String
         get() = store.string(TOKEN_KEY) ?: NSUUID().UUIDString().also {
             store.set(TOKEN_KEY, it)
```
crumpf, line 10: It looks like this `set()` returns a bool for success/failure. If the set fails would it be worthwhile to log the failure?

bjdupuis (approved):

transport/src/iosMain/kotlin/com/genesys/cloud/messenger/transport/util/TokenStoreImpl.kt
AfanasievAnton: PR updated.

transport/src/iosMain/kotlin/com/genesys/cloud/messenger/transport/util/MobileMessenger.kt
```diff
@@ -19,7 +19,7 @@ object MobileMessenger {
         val log = Log(configuration.logging, LogTag.MESSAGING_CLIENT)
         val api = WebMessagingApi(log.withTag(LogTag.API), configuration)
         val webSocket = PlatformSocket(log.withTag(LogTag.WEBSOCKET), configuration, 300000)
-        val token = TokenStoreImpl(configuration.tokenStoreKey).token
+        val token = TokenStoreImpl("configuration.tokenStoreKey23", log.withTag(LogTag.TOKEN_STORE)).token
```
crumpf, line 22: It looks like `storeKey` param is using a string literal `"configuration.tokenStoreKey23"` now

AfanasievAnton: OMG!!! What a great catch. I do not believe that I forgot to undo that change (did it to check that log.e actually work)

crumpf (approved):

PR #14, opened 2021-11-30
Addresses MMSDK-175: Build XCFramework via gradle

crumpf:
This PR retires the make-IOS-framework.sh script that had been used to create the iOS XCFramework in favor of the gradle plugin (https://kotlinlang.org/docs/mpp-build-native-binaries.html#build-xcframeworks) and can now be build by a task such as: `./gradlew :transport:assembleMessengerTransportReleaseXCFramework`


Jenkinsfile
```diff
@@ -23,10 +23,15 @@ pipeline{
     environment {
         DEPLOYMENT_ID = credentials("messenger-mobile-sdk-deployment-id")
         DEPLOYMENT_DOMAIN = 'inindca.com'
+        HOME = """${sh(
+            returnStdout: true,
+            script: 'if [ -z "$HOME" ]; then echo "/Users/$(whoami)"; else echo "$HOME"; fi'
+        ).trim()}"""
```
bjdupuis, line 29: 🙄

bjdupuis (changes requested):
Please retarget to `develop`.

Jenkinsfile
crumpf: I don't know why there's no automatic HOME env var. But, gradle touching iOS related tasks barfs without a proper HOME. I'm no Jenkins wiz, so I got past the errors by making sure a HOME is set if it doesn't already exist. I'm certainly open to suggestions on how we can make that better.

bjdupuis: My comment was only commiseration that you had to resort to such silliness.

bjdupuis (approved):

AfanasievAnton (approved):

PR #15, opened 2021-12-07
Feature/mmsdk 159

AfanasievAnton:
- Implement Reconnection mechanism to allow MessagingClient reconnect to the websocket when fatal error detected. 
This mechanism is exhaustive by default and when attempts to reconnect exceed max allowed MessagingClient will move into State.Error and halt any new attempt to reconnect.

- Add `MessagingClient.startSessionWithHistory()` API. The idea is to unite commonly used connect -> configure -> fetchNextPage calls under single function.

- Add `maxReconnectAttempts` to Configuration to allow UI ability customize amount of reconnection attempts. The default is 30. Setting it to 0 will result in immediate State.Error without any attempt to reconnect. 

- Add unit tests.

 - Add dependency to `kotlinx-coroutines-android:1.4.3-native-mt`. Required for proper unit testing of Android coroutines with KMM.

### DIFF
--- a/androidComposePrototype/src/main/java/com/genesys/cloud/messenger/androidcomposeprototype/ui/testbed/Composables.kt
+++ b/androidComposePrototype/src/main/java/com/genesys/cloud/messenger/androidcomposeprototype/ui/testbed/Composables.kt
@@ -59,7 +59,7 @@ fun TestBedContent(
             style = typography.h5
         )
         Text(
-            "Commands: connect, configure, send <msg>, history, attach, detach, delete <attachmentId> , deployment, bye, ping",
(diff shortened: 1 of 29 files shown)


bjdupuis:
FYI, a check of dependencies shows that `androidx` has _not_ crept back in. Unsure whether you explicitly checked @AfanasievAnton but let's ensure we add that as a pre-PR check.

AfanasievAnton:
> 



> FYI, a check of dependencies shows that `androidx` has _not_ crept back in. Unsure whether you explicitly checked @AfanasievAnton but let's ensure we add that as a pre-PR check.

Right now, I am manually checking for `androidx` dependencies only when I add  new one. But that is a great  idea to automate this process. Will handle that in separate pr.

bjdupuis (approved):
Let's ensure the reconnection edges are thoroughly tested.

PR #16, opened 2021-12-10
Enabling bitcode for release builds

crumpf:
We're no longer using the old `packForXcode` task which we had passed a configuration to that determined if bitcode was embedded or just marked. As of Kotlin 1.3.20, embedding bitcode is done by default for iOS framework targets. The  tasks from the XCFramework plugin appears to be building them correctly. To verify, I ran `:transport:assembleMessengerTransportReleaseXCFramework`. Then, using the recommendation from this SO (https://stackoverflow.com/questions/30722606/what-does-enable-bitcode-do-in-xcode-7/66845532#66845532), I used `otool -v -s __LLVM __bundle "transport/build/XCFrameworks/release/MessengerTransport.xcframework/ios-arm64/MessengerTransport.framework/MessengerTransport" ` to verity that the release ios-arm64 framework embedded bitcode.

### DIFF
--- a/transport/build.gradle.kts
+++ b/transport/build.gradle.kts
@@ -40,11 +40,8 @@ android {
 }
 
 val kermitVersion = "0.1.9"
-val configuration: String? by project
-val sdk: String? by project
-val bitcode: String = if ("release".equals(configuration, true)) "bitcode" else "marker"
-version = project.rootProject.version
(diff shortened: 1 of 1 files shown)


bjdupuis (approved):
I understood some of those words.

AfanasievAnton (approved):

PR #18, opened 2021-12-17
Gmdx 162 - MessagingClinet.kt async api should throw more general Exception.

AfanasievAnton:
- Use MessagingClient.fetchHistory() and MessagingClient.deploymentConfig() will throw Exception::class instead of long list of specific Exceptions.


bjdupuis (approved):

crumpf (approved):

crumpf (changes requested):
I noticed that the Jenkins build for this PR failed at the lint step due to the CancellationException import being unused.

AfanasievAnton:
Usually I run lintKotlin before pushing. This time forgot about that. :(

crumpf (approved):

PR #19, opened 2021-12-17
Update logback version to 1.2.8

crumpf:
Snyk reported a medium vulnerability in the logback dependency at version 1.2.3 which looks like it can be remediated by upgrading to at least version 1.2.7.


AfanasievAnton:
> logback

probably, because of the recent log4j vulnerability issue. It is a log hunting seasOn.

bjdupuis (approved):

AfanasievAnton (approved):

PR #20, opened 2021-12-20
GMDX-158 Document MessagingClient.State

AfanasievAnton:
- Document MessagingClient.State

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/MessagingClient.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/MessagingClient.kt
@@ -10,15 +10,48 @@ import com.genesys.cloud.messenger.transport.util.ErrorCode
 interface MessagingClient {
 
     /**
-     * MessagingClient state.
+     * Container that holds all possible MessagingClient states.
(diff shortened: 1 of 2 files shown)


JasonLPollack (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/MessagingClient.kt
```diff
@@ -10,15 +10,48 @@ import com.genesys.cloud.messenger.transport.util.ErrorCode
...
+        /**
+         * Session was successfully configured.
+         *
+         * @property connected true if configuration were established and session open.
+         * @property newSession indicates if configured session is new. When configuring an existing session [newSession] will be false.
+         */
         data class Configured(val connected: Boolean, val newSession: Boolean?) : State()
+
+        /**
+         * Invoked when the remote peer has indicated that no more incoming messages will be transmitted.
```
crumpf, line 40: Instead of `Invoked when …` which makes it seem a bit like this is a listener, how about `State when …`? (same comment applies to the `Closed` state below.

crumpf (approved):

PR #30, opened 2022-01-13
Removes unused `generateDownloadUrl` client method

crumpf:
This commit removes the `generateDownloadUrl` method from the client API since it currently does not perform any work. Further investigate with the backend team can be done to determine if this endpoint is of utility to mobile apps.

### DIFF
--- a/iosApp/iosApp/ContentViewController.swift
+++ b/iosApp/iosApp/ContentViewController.swift
@@ -173,7 +173,7 @@ class ContentViewController: UIViewController {
                 print("Socket <error>. code: <\(error.code.description)> , message: <\(error.message ?? "No message")>")
                 self?.info.text = "Socket <error>. code: <\(error.code.description)> , message: <\(error.message ?? "No message")>"
             default:
-                break
+                print("Unexpected stateListener state: \(state)")
             }
         }
     }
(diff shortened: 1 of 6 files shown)


JasonLPollack:
Is there no use-case for a customer downloading an attachment provided by an agent?

crumpf:
> Is there no use-case for a customer downloading an attachment provided by an agent?

We're going to open a follow-up ticket and discuss with the backend team to figure out what, if anything, is needed for the mobile use case and this particular endpoint.

iosApp/iosApp/ContentViewController.swift
```diff
@@ -173,7 +173,7 @@ class ContentViewController: UIViewController {
                 print("Socket <error>. code: <\(error.code.description)> , message: <\(error.message ?? "No message")>")
                 self?.info.text = "Socket <error>. code: <\(error.code.description)> , message: <\(error.message ?? "No message")>"
             default:
-                break
+                print("Unexpected stateListener state: \(state)")
```
AfanasievAnton, line 176: Side note. 
Unfortunately Kotlin\Native compiler is unable to convert sealed classes to be 'exhausted` with Swift switch statements like it does for Kotlin\Jvm.

AfanasievAnton (approved):

PR #34, opened 2022-01-19
Addresses GMDX-178

bjdupuis:
Script to perform Snyk inspection of Android transport SDK.


bjdupuis:
I've removed the environmental variable generalization attempt because I could divine no way to not explicitly spell out the variables that Snyk needs.

bjdupuis:
Closing while testing iOS changes so I don't annoy the reviewers.

PR #38, opened 2022-02-04
GMDX-209 Use new syntax to send attachments to Shyrka.

AfanasievAnton:
- Use List<Content> to send attachments via Shyrka.
- Exclude Attachment.filename and Attachment.state fields from serialization.
- Update unit tests.
- Update Podfile.lock and transport.podspec versions.

### DIFF
--- a/iosApp/Podfile.lock
+++ b/iosApp/Podfile.lock
@@ -1,6 +1,6 @@
 PODS:
   - jetfire (0.1.5)
-  - transport (1.1.0):
+  - transport (1.1.14):
     - jetfire (~> 0.1.5)
 
 DEPENDENCIES:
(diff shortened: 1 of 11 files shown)


bjdupuis:
Looks good. @AfanasievAnton I assume this is all internal to the library (and doesn't require changes to any existing users' code), correct? That's what it looks like but just asking explicitly.

AfanasievAnton:
> Looks good. @AfanasievAnton I assume this is all internal to the library (and doesn't require changes to any existing users' code), correct? That's what it looks like but just asking explicitly.

Correct. All publicly accessed objects remained unchanged. Except of Attachment.kt - but this changes are not breaking, as all that does is eliminating the `State` and `fileName` fields from serialization and giving them a default (fallback) values.
Also, attachment API is not consumed by anyone yet.

bjdupuis (approved):

crumpf (approved):

PR #40, opened 2022-02-10
build: add generateGenesysCloudMessengerTransportPodspec task 

crumpf:
Since the transport library version is now in build.gradle.kts, seemed like a decent idea to create a gradle task that can also generate the Cocoapods podspec file to be used for publication. This PR adds a new gradle task`:transport:generateGenesysCloudMessengerTransportPodspec` for that purpose. I also added the MIT License text to the podspec which removes the warning when linting/publishing the Pod.

### DIFF
--- a/transport/build.gradle.kts
+++ b/transport/build.gradle.kts
@@ -148,6 +148,16 @@ tasks {
         archiveClassifier.set("javadoc")
         from("./deployment")
     }
+
+    register("generateGenesysCloudMessengerTransportPodspec") {
+        doLast {
+            val podspecFileName = "GenesysCloudMessengerTransport.podspec"
+            var content = file("${podspecFileName}_template").readText()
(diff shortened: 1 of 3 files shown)


transport/build.gradle.kts
```diff
@@ -148,6 +148,16 @@ tasks {
         archiveClassifier.set("javadoc")
         from("./deployment")
     }
+
+    register("generateGenesysCloudMessengerTransportPodspec") {
+        doLast {
```
AfanasievAnton, line 153: Really minor. 
It looks like this task belongs to group: `group = "publishing"`

AfanasievAnton (approved):

transport/build.gradle.kts
crumpf: > Really minor.
> It looks like this task belongs to group: `group = "publishing"`

I'll add that group to the Task. While I'm at it, I'll add a description for the Task too.

PR #41, opened 2022-02-15
Feature/gmdx 227

AfanasievAnton:
### Additions:
- Abstract AttachmentHandler.
-    Add Attachment.State.Deleting - when  Attachment was requested to be deleted from the conversation history, but there were no confirmation of deletion or failure yet.
- Add Attachment.State.Sending when attachment was sent along with the parent message, but there were no confirmation of success or error yet.
- Update unit tests.
### Bug Fixes:
- Fix Bug with ios TestBed app crashing when trying to display the size of the loaded asset.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/IAttachmentHandler.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/IAttachmentHandler.kt
@@ -29,4 +29,8 @@ internal interface IAttachmentHandler {
     fun onSending()
 
     fun onSent(attachments: Map<String, Attachment>)
+
+    fun onMessageError(code: ErrorCode, message: String?)
+
+    fun clearAll()
 }
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/IAttachmentHandler.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/IAttachmentHandler.kt
@@ -28,5 +28,5 @@ internal interface IAttachmentHandler {
 
     fun onSending()
 
-    fun onSent()
+    fun onSent(attachments: Map<String, Attachment>)
 }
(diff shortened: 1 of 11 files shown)


bjdupuis:
I assume again that this is an internal change with no API contract changes? I don't see any but just want to verify.

AfanasievAnton:
No

> I assume again that this is an internal change with no API contract changes? I don't see any but just want to verify.

No breaking changes here. Addition of 2 new states for better handling of edge cases. All the rest are internal changes not visible to the consumer.

bjdupuis (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/IAttachmentHandler.kt
```diff
@@ -0,0 +1,36 @@
+package com.genesys.cloud.messenger.transport.core
+
+import com.genesys.cloud.messenger.transport.shyrka.receive.PresignedUrlResponse
+import com.genesys.cloud.messenger.transport.shyrka.receive.UploadSuccessEvent
+import com.genesys.cloud.messenger.transport.shyrka.send.DeleteAttachmentRequest
+import com.genesys.cloud.messenger.transport.shyrka.send.OnAttachmentRequest
+
+internal interface IAttachmentHandler {
```
crumpf, line 8: I don't think we've used this pattern of prefixing interfaces with `I`. Is it OK to simply call this `AttachmentHandler`?

AfanasievAnton: @crumpf  good point. 
It is more than OK. That was my initial approach, but then I realize that the `diff` this naming produce makes it difficult to see the changes I made on the implementation class. I guess this is not a good enough reason. I will follow your suggestion and rename to AttachmentHandler/AttachmentHandlerImpl .

crumpf (approved):

PR #44, opened 2022-02-25
Feature/gmdx 259 convert timestamp to unix

AfanasievAnton:
- Add kotlinx.datetime library dependency to handle iso/unix conversation for multiplatform.
- Convert Message.timestamp from iso to unix. (breaking change as it modify the structure of Message.kt)
- Add unit tests to cover conversion.
- Update existing unit tests to support changes.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/Message.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/Message.kt
@@ -10,7 +10,7 @@ import kotlinx.serialization.Serializable
  *  @property direction is direction in which message was sent.
  *  @property state represents current message state.
  *  @property type is message type. With default type "Text".
- *  @property text optional text payload of the message
+ *  @property text optional text payload of the message.
  *  @property timeStamp optional timeStamp the time when the message occurred represented in Unix epoch time, the number of milliseconds since January 1, 1970 UTC.
  *  @property attachments map of [Attachment] files to the message. Empty by default.
  */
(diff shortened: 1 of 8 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/Message.kt
```diff
@@ -13,7 +13,7 @@ data class Message(
     val state: State = State.Idle,
     val type: String = "Text",
     val text: String? = null,
-    val timeStamp: String? = null,
+    val timeStamp: Long? = null,
```
crumpf, line 16: most of the `Message` properties seem self-explanatory, but with the change of `timeStamp` from `String` (which would be an ISO timestamp) to `Long` maybe it's a good idea to document it. Maybe something like:

`*  @param timeStamp the time when the message occurred represented in Unix epoch time, the number of milliseconds since January 1, 1970 UTC.`

crumpf: Could I suggest slight wording changes to the KDoc comments? I'm also not sure if we need to call out in the param description that something is optional as it seems like that's already covered by the object type. But, I could be swayed if there is prior art for doing so.

```
  *  @property id a unique message identifier.
  *  @property direction the direction in which the message was sent.
  *  @property state the current message state.
  *  @property type the message type. The default type is "Text".
  *  @property text the text payload of the message.
  *  @property timeStamp the time when the message occurred represented in Unix epoch time, the number of milliseconds since January 1, 1970 UTC.
  *  @property attachments a map of [Attachment] files to the message. Empty by default.
```

bjdupuis (approved):

PR #46, opened 2022-03-04
Feature/gmdx 262 invert conversation order

AfanasievAnton:
- Invert message order in conversation from `new-old` to `old-new`.
- Update unit tests to reflect changes.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessageStore.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessageStore.kt
@@ -25,13 +25,17 @@ internal class MessageStore(
     fun prepareMessage(text: String): OnMessageRequest {
         val messageToSend = pendingMessage.copy(text = text, state = Message.State.Sending).also {
             log.i { "Message prepared to send: $it" }
-            activeConversation.add(0, it)
+            activeConversation.add(it)
             messageListener?.invoke(MessageEvent.MessageInserted(it))
             pendingMessage = Message()
         }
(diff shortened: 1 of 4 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessageStore.kt
```diff
@@ -25,13 +25,17 @@ internal class MessageStore(
...
             pendingMessage = Message()
         }
         return OnMessageRequest(
             token = token,
-            message = TextMessage(text, metadata = mapOf("customMessageId" to messageToSend.id), content = messageToSend.getUploadedAttachments()),
+            message = TextMessage(
+                text,
+                metadata = mapOf("customMessageId" to messageToSend.id),
+                content = messageToSend.getUploadedAttachments()
+            ),
```
crumpf, line 38: I don't think the trailing comma is needed here

AfanasievAnton: @crumpf you are correct. It is not one of recommended use cases for trailing comma. 
https://kotlinlang.org/docs/coding-conventions.html#trailing-commas

crumpf: Thanks for that link! TIL: "The Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it at your discretion for the call site." 🤯

bjdupuis (approved):

PR #51, opened 2022-04-20
GMDX-364

AfanasievAnton:
- Set timeout on REST requests for both iOS and Android platforms to 30 seconds.


bjdupuis (approved):

crumpf:
Does it make sense to this make the timeout configurable (with the default to 30 seconds) so that the user of the SDK could alter it if desired?

AfanasievAnton:
> Does it make sense to this make the timeout configurable (with the default to 30 seconds) so that the user of the SDK could alter it if desired?

It should not be a big effort to do it. But I am not sure if at this point we should overload Configuration object without actual need. Unfortunately kotlin/objc interop does not support default arguments, so if on Android this argument can be "defaulted" during Config initialization - on iOS we will have to explicitly specify values.

crumpf (approved):

crumpf:
> > Does it make sense to this make the timeout configurable (with the default to 30 seconds) so that the user of the SDK could alter it if desired?
> 
> It should not be a big effort to do it. But I am not sure if at this point we should overload Configuration object without actual need. Unfortunately kotlin/objc interop does not support default arguments, so if on Android this argument can be "defaulted" during Config initialization - on iOS we will have to explicitly specify values.

👍

PR #54, opened 2022-04-26
GMDX-376

AfanasievAnton:
- Notify Ui with empty message list when `startOfConversation = true` and `fetchNextPage()` executed. This in order to indicate end of fetchNextPage operation in cases when there are NO more history to fetch.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
@@ -147,8 +147,8 @@ internal class MessagingClientImpl(
     override suspend fun fetchNextPage() {
         checkConfigured()
         if (messageStore.startOfConversation) {
-            log.i { "All history have been fetched." }
-            throw UnsupportedOperationException("All history have been fetched.")
+            log.i { "All history has been fetched." }
+            throw UnsupportedOperationException("All history has been fetched.")
         }
         log.i { "fetching history for page index = ${messageStore.nextPage}" }
         jwtHandler.withJwt { jwt -> api.getMessages(jwt, messageStore.nextPage) }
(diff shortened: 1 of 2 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -148,7 +148,7 @@ internal class MessagingClientImpl(
         checkConfigured()
         if (messageStore.startOfConversation) {
             log.i { "All history have been fetched." }
-            return
+            throw UnsupportedOperationException("All history have been fetched.")
```
JasonLPollack, line 151: I'm not a fan of throwing in cases like this--It's not really an exceptional situation, is it?
(BTW If you do decide to throw, the language should be "All history has been fetched.")

AfanasievAnton: Other options were: 
Add boolean return type to the function with indicator of success/fail. But this has a downside of changing public api. Also it sounds a little bit unapropriate.
Second option would be simply allow network request go through and this will result in empty history response, but this also sounds wrong, as I wanted to avoid unnecessary network requests.

JasonLPollack: Does adding an exception here not change the public api?

AfanasievAnton: No. This function is already defined as throwable, therefore anybody who use it aware that exception can be thrown.

JasonLPollack: Ah, ok. That's fair.

JasonLPollack: I like this very much.

crumpf: I like this too.

crumpf (approved):

JasonLPollack (approved):

PR #59, opened 2022-05-20
GMDX-385 remove Region enum

crumpf:
Removing the Region enum decouples the Transport SDK from needing to attempt to stay in sync with the list of supported domains that continues to change as as new environments are added to Genesys Cloud. The Transport `Configuration` is already decoupled from `Region` since it it simply uses a string representation for the `domain` argument. This also allows us to drop the region generation script from the code base which is appropriate since there's no public service discovery mechanism and this is an open source project.

I also diddled around with a little wording on some of the code documentation for the client interface.


AfanasievAnton (approved):
Documentation looks much better!!!

JasonLPollack (approved):

bjdupuis (approved):

PR #61, opened 2022-06-06
GMDX-481 Create End to End tests for transport layer.

mmorehou:
Start of end to end tests that use transport SDK components to verify that we can connect and interact with Genesys Cloud. Tests are successfully configured to verify that we can send/receive messages and that we can attach and send an image. These tests are used with the iOS components of the sdk.


Jenkinsfile
```diff
@@ -85,7 +85,12 @@ pipeline{
                       echo "deploymentId=${DEPLOYMENT_ID}" >> deployment.properties
                       echo "deploymentDomain=${DEPLOYMENT_DOMAIN}" >> deployment.properties
                     fi
+			./gradlew -p "transport" :transport:syncFramework \
```
crumpf, line 88: nitpick: the indentation looks weird on these lines

iosApp/iosAppTests/Support/PublicAPI/PresenceApi.swift
```diff
@@ -0,0 +1,98 @@
...
+//  PresenceApi.swift
+//  MessengerUITests
+//
+//  Created by Morehouse, Matthew on 6/3/21.
+//  Copyright © 2021 Genesys. All rights reserved.
+//
+
+import XCTest
+
+extension ApiHelper {
```
AfanasievAnton, line 11: Do we need PresenceAPI? I am not aware that transport has any means of supporting that feature. At least, no in observable universe.

mmorehou: Presence API lets us set agents to be On Queue and will let us respond to messages that are sent from the transport SDK. So as far as the sdk goes, no the presence API is not needed. But testing needs to be able to set agents to on queue to make sure everything is working as expected.

betsyknoke (approved):

crumpf (approved):

PR #62, opened 2022-06-09
GMDX-419-Custom Attributes

AfanasievAnton:
Requirements:
As a Developer I can attach custom attributes to inbound messages to further personalized routing logic and agent management.

- Developers can pass custom attributes on every incoming message
- Can be used downstream in Architect Flows as Participant Data

In this pr:

- Add optional `customAttributes` to MessagingClient.sendMessage()
- Add Channel with Metadata when send OnMessageRequest.kt
- Add and handle ErrorCode.CustomAttributeSizeTooLarge.
- Update Android and iOS test applications to reflect the changes.
- Provide unit tests to cover new functionality.
- Remove duplicated HttpClientProvider.kt, DeploymentConfigUseCase.kt and DeploymentConfigUseCaseTest.kt


iosApp/iosApp/ContentViewController.swift
```diff
@@ -322,6 +325,15 @@ extension ContentViewController : UITextFieldDelegate {
             })
         case ("clearConversation", _):
             client.invalidateConversationCache()
+        case ("addAttribute", let msg?):
+            let segments = splitUserInput(msg)
+            if((segments.0?.isEmpty) == nil) {
```
crumpf, line 330: if `segments.0` is an empty string here this will still fall through to the `else` clause. I'd recommend restructuring like the following:

```
case ("addAttribute", let msg?):
    let segments = splitUserInput(msg)
    if let key = segments.0, !key.isEmpty {
        customAttributes[key] = segments.1 // is it ok for segments.1 (the value) to be nil or should it be checks in the if clause too?
        self.info.text = "Custom attribute added: \(segments)"
    }  else {
        self.info.text = "Custom attribute key cannot be null or empty!"
    }
```

AfanasievAnton: @crumpf Thanks for feedback. 
I have updated the code. And no `segment. 1` should not be nil. So I added a check to fallback to empty string if it is.

iosApp/iosApp/ContentViewController.swift
```diff
@@ -237,17 +247,18 @@ extension ContentViewController : UITextFieldDelegate {
 
         attachImageName = ""
         let userInput = segmentUserInput(message)
-
+        let command = convertToCommand(command: userInput.0,input: userInput.1)
```
crumpf, line 250: I have some ideas on how to refactor some of this user input handling, but I think I'll tackle that in a separate PR in the future.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
```diff
@@ -96,10 +96,11 @@ interface MessagingClient {
      * Send a message to the conversation as plain text.
      *
      * @param text the plain text to send.
+     * @param customAttributes optional dictionary of attributes to send with the message. Empty by default.
      * @throws IllegalStateException
      */
     @Throws(IllegalStateException::class)
-    fun sendMessage(text: String)
+    fun sendMessage(text: String, customAttributes: Map<String, String> = emptyMap())
```
crumpf, line 103: Since default params cannot bridge to ObjC, what do you think about having two methods so that iOS isn't always required to supply `customAttributes` if they're not needed.
`fun sendMessage(text:)` and `fun sendMessage(text:customAttributes)`?

AfanasievAnton: That is interesting comment. I remember having an issue with obj-c and default arguments. However it seems not to be a case anymore. Based on how MessengerTransport.h looks like, it is capable of doing that:
``` func sendMessage(text: String, customAttributes: [String : String] = [:]) throws ``` 
Which looks legit to me.
I am looking for additional information to confirm that, and will update here if I find anything.

AfanasievAnton: Update:

"An argument whose Objective-C type is an NSDictionary defaults to nil if the NSDictionary is nullable and [:] if the NSDictionary is non-nullable, under the following conditions:

* the argument label contains the word "options" or "attributes", or the two words "user info" one after another
* the argument label is empty and the method base name ends with the word "options" or "attributes", or the two words "user info" one after another"

https://github.com/apple/swift/blob/main/docs/CToSwiftNameTranslation.md#default-argument-values


As a conclusion it is technically ok to have customAttributes as default argument. 
Having `customAttributes` as additional argument on `sendMessage()` feels more correct design-wise. Based on their requirement they should not be stored anywhere, instead they should be passed on every incoming message.

crumpf (approved):

PR #68, opened 2022-06-29
Feature/gmdx 527

AfanasievAnton:
- Use parametrized connect(shouldConfigure: Boolean) to prepare for deprecation of connect/configure.
- Deprecate connect and configure. In the next major release they will become internal.
- Add connectWithConfigure command to the testbed application on ios and Android.
- Provide unit test coverage for newly added api.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
@@ -91,7 +91,7 @@ interface MessagingClient {
      * Open a secure WebSocket connection to the Web Messaging service with the url and
      * deploymentId configured on this MessagingClient instance.
      *
-     * @param shouldConfigure when true, Configure a Web Messaging session.
+     * @param shouldConfigure a Boolean value indicating whether to configure a Web Messaging session after the WebSocket is opened.
      * @throws IllegalStateException
      */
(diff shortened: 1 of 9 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
```diff
@@ -83,12 +83,24 @@ interface MessagingClient {
...
 
+    /**
+     * Open a secure WebSocket connection to the Web Messaging service with the url and
+     * deploymentId configured on this MessagingClient instance.
+     *
+     * @param shouldConfigure when true, Configure a Web Messaging session.
+     * @throws IllegalStateException
+     */
+    @Throws(IllegalStateException::class)
+    fun connect(shouldConfigure: Boolean = false)
```
crumpf, line 98: Just wondering if `shouldConfigure` should default to `true` instead of `false`.

Also wondering about the parameterization of configuring the session in this connect method. Is the idea to be able to preserve the current functionality in the future, when `connect()` and `configureSession()` are removed. If so and the SDK user wanted the same behavior today being able to separate socket connection and session configuration, would they use this interface as below?

```
// if trying to keep behavior in future same as today, does this:
connect()
// ...
configureSession()

// become this:

connect(shouldConfigure = false)
// ...
connect(shouldConfigure = true)
```

mmorehou (approved):

PR #70, opened 2022-07-14
GMDX-528-Reconnect

AfanasievAnton:
- Add ping-based network monitoring on iOS.
- Pings will be sent every 20 seconds.
- Expected interval for receiving a pong is 10 seconds.
- Failure to receive a pong in specified interval will cause websocket failure.
- Slightly modify PlatformSocket.kt constructor to use kotlin default values.
- Prevent cal to listener.onOpen() on iOS if ws is null (edge case)
- Reorder functions in PlatformSocket.kt (iOS) for better readability.
- Add kdocs.

### DIFF
--- a/transport/src/iosMain/kotlin/com/genesys/cloud/messenger/transport/network/PlatformSocket.kt
+++ b/transport/src/iosMain/kotlin/com/genesys/cloud/messenger/transport/network/PlatformSocket.kt
@@ -1,9 +1,9 @@
 package com.genesys.cloud.messenger.transport.network
 
-import com.genesys.cloud.messenger.transport.core.Configuration
 import com.genesys.cloud.messenger.transport.util.extensions.string
 import com.genesys.cloud.messenger.transport.util.extensions.toNSData
 import com.genesys.cloud.messenger.transport.util.logs.Log
+import io.ktor.http.Url
 import kotlinx.cinterop.convert
 import platform.Foundation.NSData
 import platform.Foundation.NSError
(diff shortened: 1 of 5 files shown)


transport/src/iosMain/kotlin/com/genesys/cloud/messenger/transport/network/PlatformSocket.kt
```diff
@@ -108,55 +125,45 @@ internal actual class PlatformSocket actual constructor(
...
-            when {
-                nsError != null -> {
-                    handleErrorAndDeactivate(nsError, "Receive handler error")
-                    return@receiveMessageWithCompletionHandler
-                }
-                message != null -> {
-                    message.string?.let { listener.onMessage(it) }
-                }
+    private fun sendPing() {
+        log.i { "Sending ping" }
```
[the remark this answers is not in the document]
AfanasievAnton: That`s a great catch

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/StateMachineTest.kt
```diff
@@ -204,6 +204,30 @@ class StateMachineTest {
...
+        subject.onSessionConfigured(connected = true, newSession = true)
+        assertThat(subject).isConfigured(connected = true, newSession = true)
+        subject.onReconnect()
+        assertThat(subject).isReconnecting()
+        subject.onConnect()
+        assertThat(subject).isReconnecting()
+        subject.onConnectionOpened()
+        assertThat(subject).isReconnecting()
+        subject.onConfiguring()
+        subject.onSessionConfigured(connected = true, newSession = true)
```
crumpf, line 223: in most cases after reconnecting, should it be `newSession = false`?

AfanasievAnton: It should, but since it is something we receive from BE it does not really important to capture it here. All we want to do is to validate that parameters are not modified and that they match the one that are passed in. 
I will update the `newSession` to `false` to keep the flow clean.

transport/src/iosMain/kotlin/com/genesys/cloud/messenger/transport/network/PlatformSocket.kt
```diff
@@ -108,55 +124,49 @@ internal actual class PlatformSocket actual constructor(
...
+        if(waitingOnPong) {
```
crumpf: <style nitpick> I think we typically have a space between the `if` and the opening `(`

AfanasievAnton: You are correct. Forgot to run the lint command before pushing.

crumpf (approved):

PR #71, opened 2022-07-23
ios websocket improvements

crumpf:
1. the "did open" and "did close" delegates now only do stuff if the `webSocket` reference matches the `webSocketTask` that's supplied in the callback. This way we're sure that we're taking action on the right websocket that PlatformSocket is currently using.
2. created `deactivateAndCancelWebSocket(code:reason:)`. This method calls `deactivate()` in order to cancel pings and release the webSocket reference, and after doing so attempts to cancel the webSocket with the given code and reason. Performing the socket close _after_ nulling `webSocket` prevents the "did close" delegate callback from repeating the deactivate and listener.onClosed work due to the change mentioned in #1.
3. refactor `handleErrorAndDeactivate(...)` to just `handleError(...)`. `handleError()` now calls `deactivateAndCancelWebSocket()` which deactivates and closes the websocket to hopefully preventing any other work from being done on it.
4. made a change in `closeSocket(code:reason)` so that it calls `deactivateAndCancelWebSocket(...)` and then immediately calls `listener?.onClosed()` instead of waiting to do it in the delegate callback. The thought here is that we're really just force closing the socket so why not release the websocket and immediately signal that we're closed?

### DIFF
--- a/transport/src/iosMain/kotlin/com/genesys/cloud/messenger/transport/network/PlatformSocket.kt
+++ b/transport/src/iosMain/kotlin/com/genesys/cloud/messenger/transport/network/PlatformSocket.kt
@@ -22,8 +22,6 @@ import platform.Foundation.setValue
 import platform.darwin.NSObject
 import platform.posix.ETIMEDOUT
 
-private const val CUSTOM_SOCKET_CLOSE_CODE = -111L
-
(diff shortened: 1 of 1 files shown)


crumpf:
The websocket idea proposed in this PR was incorporated into PR #72. Therefore, I'm closing this PR.

PR #74, opened 2022-07-26
GMDX-528-Reconnect

AfanasievAnton:
- Add ErrorCode.NetworkDisabled. This error coded will notify UI that user has disabled network and has to enable it in order to proceed. This will also set Transport to State.Error immediately.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
@@ -210,19 +210,22 @@ internal class MessagingClientImpl(
                     stateMachine.onError(code, message)
                 }
             }
-            is ErrorCode.WebsocketError -> handleWebSocketError()
+            is ErrorCode.WebsocketError -> handleWebSocketError(
+                ErrorCode.WebsocketError,
+                ErrorMessage.FailedToReconnect
+            )
             else -> log.w { "Unhandled ErrorCode: $code with optional message: $message" }
         }
     }
 
-    private fun handleWebSocketError() {
+    private fun handleWebSocketError(errorCode: ErrorCode, message: String? = null) {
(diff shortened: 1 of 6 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -210,19 +210,22 @@ internal class MessagingClientImpl(
...
+            )
             else -> log.w { "Unhandled ErrorCode: $code with optional message: $message" }
         }
     }
 
-    private fun handleWebSocketError() {
+    private fun handleWebSocketError(errorCode: ErrorCode, message: String? = null) {
         if (stateMachine.isClosed()) return
         invalidateConversationCache()
         if (reconnectionHandler.shouldReconnect) {
```
crumpf, line 224: If the intention is to immediately notify UI that the network has been disabled, we should check to see if the `errorCode` is `NetworkDisabled` and doing something appropriate before running through the current `if … else …` logic. As it currently stands in the event of the socket reporting `NetworkDisabled` it will still rapidly run through the reconnectionHandler until maxAttempts is reached and then end up in the final error state.

Maybe we could add a `invalidate` method to the ReconnectionHandler so if we get network disabled error we can "turn off" the reconnection handler before check to see if to reconnect? e.g.: 

```
if (error == NetworkDisabled)  { reconnectionHandler.invalidate() }

if (reconnectionHandler.shouldReconnect) {
    stateMachine.onReconnect()
    reconnectionHandler.reconnect { connect(true) }
} else {
    stateMachine.onError(errorCode, message)
    attachmentHandler.clearAll()
}
```

AfanasievAnton: Good idea. I think I need to refactor this part a little bit.

crumpf (approved):

betsyknoke (approved):

PR #89, opened 2022-08-22
MTSDK-26 Send Typing Indicator.

AfanasievAnton:
- Add `userIsTyping` public api to send `typing` events from user.
- Ensure that `typing` indicator not sent more that once every 5 seconds.
- Extract commonality between EventMessage.kt and TextMessage.kt to BaseMessageProtocol.kt
- Add EventMessage.kt data class for `events` sent to Shyrka.
- Ensure that UI is notified about Outbound events only.
- Include eventType in StructuredMessageEvent.kt
- Just to be on the safe side, use type Long for `duration` field instead of Int.
- Add/Update unit tests.
- Add KDocs for public API.

### DIFF
--- a/transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/util/Request.kt
+++ b/transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/util/Request.kt
@@ -1,0 +1,8 @@
+package com.genesys.cloud.messenger.transport.util
+
+internal object Request {
+    const val configureRequest =
+        """{"token":"00000000-0000-0000-0000-000000000000","deploymentId":"deploymentId","journeyContext":{"customer":{"id":"00000000-0000-0000-0000-000000000000","idType":"cookie"},"customerSession":{"id":"","type":"web"}},"action":"configureSession"}"""
+    const val userTypingRequest =
+        """{"token":"00000000-0000-0000-0000-000000000000","action":"onMessage","message":{"events":[{"eventType":"Typing","typing":{"type":"On","duration":5000}}],"type":"Event"}}"""
+}
--- a/transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/util/Request.kt
+++ b/transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/util/Request.kt
@@ -4,5 +4,5 @@ internal object Request {
     const val configureRequest =
         """{"token":"00000000-0000-0000-0000-000000000000","deploymentId":"deploymentId","journeyContext":{"customer":{"id":"00000000-0000-0000-0000-000000000000","idType":"cookie"},"customerSession":{"id":"","type":"web"}},"action":"configureSession"}"""
     const val userTypingRequest =
-        """{"token":"00000000-0000-0000-0000-000000000000","action":"onMessage","message":{"events":[{"eventType":"Typing","typing":{"type":"On","duration":5000}}],"type":"Event"}}"""
+        """{"token":"00000000-0000-0000-0000-000000000000","action":"onMessage","message":{"events":[{"eventType":"Typing","typing":{"type":"On"}}],"type":"Event"}}"""
 }
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
@@ -142,7 +142,7 @@ interface MessagingClient {
 
     /**
      * Attach a file to the message. This file will be uploaded and cached locally
-     * until user decides to send a message.
+     * until customer decides to send a message.
      * After the message has been sent, attachment will be cleared from cache.
      *
      * @param byteArray data to upload.
@@ -193,10 +193,12 @@ interface MessagingClient {
     fun invalidateConversationCache()
 
     /**
-     * Notify the Agent that the User is typing a message.
+     * Notify the agent that the customer is typing a message.
+     * This command sends a single typing indicator event and should be called a maximum of once every 5 seconds.
+     * If called more frequently, this command will be rate limited in order to optimize network traffic.
      *
      * @throws IllegalStateException if called before session was connected.
      */
     @Throws(IllegalStateException::class)
-    fun userIsTyping()
+    fun indicateTyping()
 }
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
@@ -191,4 +191,12 @@ interface MessagingClient {
      * latest available history.
      */
     fun invalidateConversationCache()
+
+    /**
+     * Notify the Agent that the User is typing a message.
+     *
+     * @throws IllegalStateException if called before session was connected.
+     */
+    @Throws(IllegalStateException::class)
+    fun userIsTyping()
 }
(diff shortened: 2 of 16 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
```diff
@@ -191,4 +191,12 @@ interface MessagingClient {
...
      */
     fun invalidateConversationCache()
+
+    /**
+     * Notify the Agent that the User is typing a message.
+     *
+     * @throws IllegalStateException if called before session was connected.
+     */
+    @Throws(IllegalStateException::class)
+    fun userIsTyping()
```
crumpf, line 201: Given my above comment, does it make sense to change this method signature? Maybe one of the following?

`fun indicateTyping()`

or

`fun sendTypingIndicator()`

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -198,6 +202,17 @@ internal class MessagingClientImpl(
         messageStore.invalidateConversationCache()
     }
 
+    @Throws(IllegalStateException::class)
+    override fun userIsTyping() {
+        val encodedJson = userTypingProvider.encodeRequest(token)
+        if (encodedJson == null) {
+            log.w { "Typing event can be sent only once every $TYPING_INDICATOR_COOL_DOWN_IN_MILLISECOND milliseconds." }
```
crumpf, line 209: If we're in this cool-down interval, maybe we should throw an error to let the caller know they were rate limited.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
```diff
@@ -191,4 +191,12 @@ interface MessagingClient {
      * latest available history.
      */
     fun invalidateConversationCache()
+
+    /**
+     * Notify the Agent that the User is typing a message.
```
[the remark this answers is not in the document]
AfanasievAnton: They use term `user` as well, but indeed - `customer` is used more often. I will update the kdocs and keep it in mind for future documentation.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
AfanasievAnton: I will change to `sendTypingIndicator`

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/util/Request.kt
```diff
@@ -0,0 +1,8 @@
+package com.genesys.cloud.messenger.transport.util
+
+internal object Request {
+    const val configureRequest =
+        """{"token":"00000000-0000-0000-0000-000000000000","deploymentId":"deploymentId","journeyContext":{"customer":{"id":"00000000-0000-0000-0000-000000000000","idType":"cookie"},"customerSession":{"id":"","type":"web"}},"action":"configureSession"}"""
+    const val userTypingRequest =
+        """{"token":"00000000-0000-0000-0000-000000000000","action":"onMessage","message":{"events":[{"eventType":"Typing","typing":{"type":"On","duration":5000}}],"type":"Event"}}"""
```
crumpf, line 7: Was it agreed by Shyrka team that the "duration" is included in the typing indicator events for both customer and agent typing? (only asking because it's not in the Developer Center documentation)

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
AfanasievAnton: There was some back and forth with myself about that, but I could not find good arguments to prefer throwing an exception on top of the logging. In such a case, throwing an exception seems excessive. In any case, if we throw an exception, the UI side doesn't have anything to do with it.
Additionally, in a potential implementation, this function might be invoked whenever an input field changes. There will be a lot of exceptions generated by that.

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/util/Request.kt
AfanasievAnton: I have raised this question in the `typing indicator` chatroom, but was ignored. I will dm Julien from Shyrka. usually he is the person who deals with their documentation and is very helpful.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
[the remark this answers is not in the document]
AfanasievAnton: Also, while changing from `user` to `customer` I realized that our backend use term `user` for typing indicator fields in deploymentConfig. So now I am a little ambiguous regarding what term should be used.
@crumpf wonder if you think that is fine to use term `customer` for KDoc and other documentation, but internally use the term `User` for cases like UserTypingProvider and similar?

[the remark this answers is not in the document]
bjdupuis: It's so tricky with these terms. `Customer` from what I've been imagining in there scenarios would be like Quicken. `End user` or just `User` would be the person using the application. Of course, if there's prior art in Web Messaging already that's tricky but it feels like we should definitely standardize.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
crumpf: Ok, I think that makes sense to me. My thought was that maybe the customer would want to know when they're being rate limited, but in practical use of this API it probably isn't needed. Like you mention, the most common use would be to just call the API on input field changes and let us just handle the rate limiting to ensure we don't send more than one typing indicator in any 5 second interval.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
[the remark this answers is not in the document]
crumpf: Ugh. @bjdupuis is right the nomenclature can be difficult. Now I'm having flashbacks to some of the user research studies we were a part of 😬. I agree that we should standardize on these terms, also with the DX mobile, Web Messenger, Messenger, et al. teams.

@AfanasievAnton I'm fine using `user` internally, especially if it helps match terms in the backend use to avoid confusion. Maybe for now, given the prior art in the Web Messaging documentation, we use "user" internally and "customer" in the docs, like you proposed, and try to keep "user" or "customer" out of the SDKs public API unless the user is ambiguous.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
crumpf: I think I have an ever so slight preference for `indicateTyping()`.

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/util/Request.kt
AfanasievAnton: [> Was it agreed by Shyrka team that the "duration" is included in the typing indicator events for both customer and agent typing? (only asking because it's not in the Developer Center documentation)]
So, after discussion with Shyrka team, it was decided that:
1. `duration` field should be added to the [agent typing response](https://bitbucket.org/inindca/developer-center/pull-requests/1131/shyrka-1627-add-duration-to-typing-event?link_source=email). There are a [pr opened](https://bitbucket.org/inindca/developer-center/pull-requests/1131/shyrka-1627-add-duration-to-typing-event?link_source=email)
2. The `"type":"On"` must be included in [customer typing request](https://developer.genesys.cloud/commdigital/digital/webmessaging/websocketapi#customer-typing).
3. `duration` should be removed from [customer typing request](https://developer.genesys.cloud/commdigital/digital/webmessaging/websocketapi#customer-typing).
4. `duration` must be an optional value in agent typing response with fallback value of 5000ms if not set. 

Going to update this pr accordingly with following changes:
1.  Remove `duration` from user typing request.
2. Make `duration` a nullable with fallback value of 5000 ms if not present.

bjdupuis (approved):

mmorehou (approved):

crumpf (approved):

AfanasievAnton:
Jenkins pipeline ran successfully.

PR #94, opened 2022-08-30
MTSDK-66 Handle typing indicator errors.

AfanasievAnton:
- Notify UI when Shyrka respond with error to sending typing indicator event.
- Add Event.Error.
- Add `correctiveAction` for unstructured errors according to the Shyrka documentation: https://developer.genesys.cloud/commdigital/digital/webmessaging/websocketapi#http-error-codes
- Add unit tests.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
@@ -13,6 +13,7 @@ import com.genesys.cloud.messenger.transport.network.SocketCloseCode
 import com.genesys.cloud.messenger.transport.network.WebMessagingApi
 import com.genesys.cloud.messenger.transport.shyrka.WebMessagingJson
 import com.genesys.cloud.messenger.transport.shyrka.receive.AttachmentDeletedResponse
+import com.genesys.cloud.messenger.transport.shyrka.receive.ErrorEvent
 import com.genesys.cloud.messenger.transport.shyrka.receive.GenerateUrlError
 import com.genesys.cloud.messenger.transport.shyrka.receive.JwtResponse
 import com.genesys.cloud.messenger.transport.shyrka.receive.PresignedUrlResponse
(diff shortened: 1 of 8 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -233,6 +234,8 @@ internal class MessagingClientImpl(
                 if (stateMachine.isConnected()) {
                     shouldConfigureAfterConnect = false
                     stateMachine.onError(code, message)
+                } else {
+                    eventHandler.onEvent(ErrorEvent(errorCode = code, message = message))
```
crumpf, line 238: Under which scenarios could we fall into this scenario? Trying to understand how we signal an `ErrorEvent` (a type of `StructuredMessageEvent`) here instead of a state error.

AfanasievAnton: Mostly regular 4xx error range. For instance: 
Toggle off the typing indicator in Admin console and try sending a `userTyping` request will return Forbidden 403.
In that case there are no reason to fall into error state, but rather notify UI with event.

crumpf: That makes sense to me, that response errors that are due to the client might not cause us to fall into the error state but instead send an error event. I guess what I'm wondering is that right now it's written so that if a `ClientReponseError` happens while connected state then we go to error state, otherwise we report `ErrorEvent`. Would we want all response errors in the client range to report `ErrorEvent`? e.g.

```
…
is ErrorCode.ClientResponseError
-> eventHandler.onEvent(ErrorEvent(errorCode = code, message = message))
is ErrorCode.ServerResponseError,
is ErrorCode.RedirectResponseError,
-> {
    if (stateMachine.isConnected()) {
        shouldConfigureAfterConnect = false
        stateMachine.onError(code, message)
    }
}
…
```

AfanasievAnton: The ClientResponseError during `State.Connected` is an edge case. It points out to that `configureSession` request has failed, and therefore we can not proceed with any action. This is why I transition MessengerClient into the State.Error.

crumpf: Oh! I was confusing `isConnected` with `isConfigured`, so that explains why I was having a hard time wrapping my head around it.

crumpf (approved):

mmorehou (approved):

betsyknoke (approved):

AfanasievAnton:
Jenkins had a green run.

PR #96, opened 2022-08-30
Feature/mtsdk 71 xcframework metadata

crumpf:
This PR adds a post action hook to the `assembleMessengerTransportXCFramework` gradle task which updates the built framework artifacts with the correct version metadata.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -97,7 +97,7 @@ pipeline{
         }
         stage("CI Build - iOS XCFramework"){
             steps{
-                sh './gradlew assembleMessengerTransportXCFramework'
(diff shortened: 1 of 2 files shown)


crumpf:
On second thought, I think it makes more sense to add the post-task hooks to `assembleMessengerTransportDebugXCFramework` and `assembleMessengerTransportReleaseXCFramework` which are used in the calling of `assembleMessengerTransportXCFramework` in case those debug or release tasks are used independently of the umbrella assembly tasks. Working on an update.

crumpf:
Jenkins ran the pipeline successfully

AfanasievAnton (approved):

mmorehou (approved):

PR #97, opened 2022-08-31
update: iOS testbed state checking to match docs

crumpf:
Updating the testbed to reflect some of the examples currently used in the Developer Center where `is` is used to check type instead of `as` down-casting to an unnamed variable.


mmorehou (approved):

AfanasievAnton (approved):

crumpf:
Jenkins pipeline ✅

PR #116, opened 2022-09-28
remove logback dependency

crumpf:
Doesn't look like we're using logback anywhere so we should be able to drop it.


crumpf:
Jenkins is happy ✅

AfanasievAnton (approved):

PR #120, opened 2022-10-13
MTSDK-96 Add Autostart event.

AfanasievAnton:
- Add AutostartRequest model to send to the Shyrka.
- Map Autostart (Presence) event to ConversationAutostart event.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/events/Event.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/events/Event.kt
@@ -32,4 +32,9 @@ sealed class Event {
         val message: String?,
         val correctiveAction: CorrectiveAction,
     ) : Event()
+
+    /**
+     * This event indicates that conversation was successfully autostart.
+     */
+    object ConversationAutostart : Event()
 }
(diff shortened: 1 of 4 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/events/Event.kt
```diff
@@ -32,4 +32,9 @@ sealed class Event {
         val message: String?,
         val correctiveAction: CorrectiveAction,
     ) : Event()
+
+    /**
+     * This event indicates that conversation was successfully autostart.
```
bjdupuis, line 37: perhaps "autostarted"?

AfanasievAnton: Does it feel correct to rename the event to "ConversationAutostarted" as well?

bjdupuis: I'm iffy on that one. It's better _English_ but I don't know if it makes sense to change it. @crumpf what say you?

crumpf: > I'm iffy on that one. It's better _English_ but I don't know if it makes sense to change it. @crumpf what say you?

I'm fine with it staying `ConversationAutostart` instead of using the -ed postfix.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/events/Event.kt
```diff
@@ -32,4 +32,9 @@ sealed class Event {
         val message: String?,
         val correctiveAction: CorrectiveAction,
     ) : Event()
+
+    /**
+     * This event indicates that conversation was successfully autostarted.
```
crumpf, line 37: "This event indicates that the conversation was successfully autostarted."

AfanasievAnton:
Jenkins had a successful build.

PR #121, opened 2022-10-18
MTSDK-97 Dispatch Autostart.

AfanasievAnton:
- Send autostart request when session new session is configured and autostart is enabled.
- Notify UI with Event.Autostart when autostart request was successfully received by Shyrka.
- Add unit test to cover newly added functionality.

### DIFF
--- a/transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImplTest.kt
+++ b/transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImplTest.kt
@@ -11,12 +11,17 @@ import com.genesys.cloud.messenger.transport.network.PlatformSocketListener
 import com.genesys.cloud.messenger.transport.network.ReconnectionHandlerImpl
 import com.genesys.cloud.messenger.transport.network.TestWebMessagingApiResponses
 import com.genesys.cloud.messenger.transport.network.WebMessagingApi
+import com.genesys.cloud.messenger.transport.shyrka.receive.Apps
+import com.genesys.cloud.messenger.transport.shyrka.receive.Conversations
 import com.genesys.cloud.messenger.transport.shyrka.receive.DeploymentConfig
 import com.genesys.cloud.messenger.transport.shyrka.receive.ErrorEvent
+import com.genesys.cloud.messenger.transport.shyrka.receive.PresenceEvent
 import com.genesys.cloud.messenger.transport.shyrka.receive.StructuredMessageEvent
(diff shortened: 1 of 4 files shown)


transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImplTest.kt
```diff
@@ -692,6 +697,86 @@ class MessagingClientImplTest {
         verify { mockEventHandler.onEvent(expectedEvent) }
     }
 
+    @Test
```
bjdupuis, line 700: 🤌

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -364,3 +383,6 @@ internal class MessagingClientImpl(
         }
     }
 }
+
+private fun KProperty0<DeploymentConfig?>.isAutostartEnabled(): Boolean =
+    this.get()?.messenger?.apps?.conversations?.autoStart?.enabled == true
```
bjdupuis, line 388: Choo choo, null check 🚞🚃🚃🚃🚃🚋 coming through.

bjdupuis (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
AfanasievAnton: Imagine that with if-not-null block.

transport/src/commonTest/kotlin/com/genesys/cloud/messenger/transport/shyrka/receive/FakeDeploymentConfig.kt
```diff
@@ -1,26 +1,23 @@
 package com.genesys.cloud.messenger.transport.shyrka.receive
 
-fun fakeDeploymentConfig() = DeploymentConfig(
+fun testDeploymentConfig(
```
crumpf, line 3: For these kinds of methods which are basically factory methods, I was wondering if it would make sense to prefix them with "create" or "make". My reasoning is that the word "test" can be a verb, so `testDeploymentConfig()` at first was ambiguous to me in knowing if it was performing some kind of test and then returning a `DeploymentConfig` or just creating a `DeploymentConfig`.

crumpf (approved):

PR #122, opened 2022-10-24
MTSDK-98 Handle Autostart failures.

AfanasievAnton:
- Wrap the deploymentConfig fetch under try/catch to avoid hard failure when request fails with an exception.


crumpf (approved):

mmorehou (approved):

AfanasievAnton:
Jenkins had a successful build.

PR #123, opened 2022-10-26
Feature/mtsdk 73 conversation autostart

AfanasievAnton:
- Add support to automatically start conversation. 
- Austostart request will be sent immediately after receiving Configure response with `newSession = true` AND`DeploymentConfig.autostart=true`.
- Upon autostart Messenger Transport will dispatch `Event.ConversationAutostart` to let UI know.
- DeploymentConfig are now cached in MessengerTransport and its getter is exposed to MessagingClient. 
- If at the moment of MessagingClient creation DeploymentConfig are null MessengerTransport will execute the fetchDeploymentConfig.
- Provide unit tests to cover newly added functionality.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
@@ -13,6 +13,7 @@ import com.genesys.cloud.messenger.transport.network.SocketCloseCode
 import com.genesys.cloud.messenger.transport.network.WebMessagingApi
 import com.genesys.cloud.messenger.transport.shyrka.WebMessagingJson
 import com.genesys.cloud.messenger.transport.shyrka.receive.AttachmentDeletedResponse
+import com.genesys.cloud.messenger.transport.shyrka.receive.DeploymentConfig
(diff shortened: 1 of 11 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -262,6 +275,12 @@ internal class MessagingClientImpl(
                     structuredMessage.events.forEach {
                         eventHandler.onEvent(it)
                     }
+                } else {
+                    structuredMessage.events.forEach {
+                        if (it.eventType == StructuredMessageEvent.Type.Presence) {
```
crumpf, line 280: This triggers `onEvent` calls for customer joins, right? (just making sure I understand)

AfanasievAnton: correct. Want to highlight that it is true for Inbound messages only, meaning that it will trigger the Event.ConversationAutostart ONLY when the Presence (autostart) was sent by customer.

crumpf (approved):

bjdupuis (approved):

PR #125, opened 2022-10-28
MTSDK-102: New automated test to verify Autostart works.

mmorehou:
* New test case to verify Autostart works as expected.
* Overriding the saved token in the auto start test to make sure we start with a new session.
* Add new conversation handling. We'll now close all existing conversations prior to tests starting.

### DIFF
--- a/iosApp/iosAppTests/Support/PublicAPI/Conversations.swift
+++ b/iosApp/iosAppTests/Support/PublicAPI/Conversations.swift
@@ -14,12 +14,22 @@ extension ApiHelper {
         let agentId = TestConfig.shared.config?.agentId ?? ""
         changePresence(presenceName: "On Queue", userID: agentId)
         guard let conversation = waitForConversation() else {
-            XCTFail("Failed to receive a new conversation to answer.")
+            print("Failed to receive a new conversation to answer.")
             return nil
         }
         sendConnectOrDisconnect(conversationInfo: conversation, connecting: true, wrapup: false)
         changePresence(presenceName: "Available", userID: agentId)
         return getConversation(conversationId: conversation.conversationId)
+    }
+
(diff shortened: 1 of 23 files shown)


AfanasievAnton (approved):

iosApp/iosAppTests/Support/PublicAPI/Conversations.swift
```diff
@@ -14,14 +14,24 @@ extension ApiHelper {
...
+        guard let conversationList = getSmsConversations()?.value(forKey: "entities") as? [JsonDictionary] else {
+            XCTFail("Failed to get the existing conversation list.")
+            return
+        }
+        for conversation in conversationList {
+            sendConnectOrDisconnect(conversationInfo: ConversationInfo(json: conversation), connecting: false)
+        }
+    }
+
     private func getSmsConversations() -> JsonDictionary? {
```
crumpf, line 35: What does the "Sms" stand for?

mmorehou: Back when this ACD type was first released, it was the "sms" media type. Now it's called the "Message" media type but the function name I use for it hasn't been updated with that just yet.

iosApp/iosAppTests/iosAppTests.swift
```diff
@@ -67,6 +73,28 @@ class iosAppTests: XCTestCase {
         contentController.disconnectMessenger()
     }
 
+    // Test Assumption: Deployment has AutoStart enabled. See Deployment Configuration options in Admin.
+    func testAutoStart() {
```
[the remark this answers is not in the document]
mmorehou: Excellent idea. I'll add a check for that.

crumpf (approved):

PR #127, opened 2022-11-03
MTSDK-15 Handle ConnectionClosedEvent

AfanasievAnton:
- Handle `ConnectionClosedEvent` and explicitly close ws when received.
- Notify UI when connection close was received with `Event.ConnectionClosed`.
- Add KDoc for newly added Event.ConnectionClosed


bjdupuis (approved):

mmorehou (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/events/Event.kt
```diff
@@ -37,4 +37,10 @@ sealed class Event {
      * This event indicates that the conversation was successfully autostarted.
      */
     object ConversationAutostart : Event()
+
+    /**
+     * Sent when connection is closing due to exceeding maximum of
+     * 3 simultaneously opened sessions.
```
crumpf, line 43: Since the number of simultaneously open session (currently 3) is enforced in the backend, it doesn't seem like we have a guarantee that this number might not change at some point in the future. That makes me wonder if we should refrain from putting that magic number in the comment.

AfanasievAnton: That is a good point. Based on my conversations with Shyrka team this value is not going to be changed anywhere in observable future. 
What do you think about like that:
 `Sent when connection is closing due to exceeding maximum number of simultaneously opened sessions. For more information about the maximum number of sessions allowed, please refer to the official documentation.` 
 
 And then specify/update allowed amount according to Shyrka. 
Or include a reference to the Shyrka documentation. 

The main problem with that is that this documentation does not exist yet. So I have nowhere to refer. There are a ticket that I am watching to track the progress on that: https://inindca.atlassian.net/browse/SHYRKA-1709

crumpf: I wonder if it's implicit that more detailed documentation about the service can be found at the Developer Center and we could just leave it at:
`Sent when the connection is closed due to exceeding the maximum number of simultaneously open sessions.`
Or if we'd want to try and point them there:
`Sent when the connection is closed due to exceeding the maximum number of simultaneously open sessions. Detailed information about Genesys Cloud Web Messaging capabilities is available in the [Developer Center.(https://developer.genesys.cloud).`

AfanasievAnton: I feel like giving a `maximum number` without saying what this number is or providing bread crumbs where to find more detailed might be confusing/frustrating. 

So I would prefer the second option, which is slightly better than what I propose as it gives a link to the documentation.

crumpf: 👍

crumpf (approved):

crumpf (approved):

AfanasievAnton:
jenkins pipeline has been fixed

PR #129, opened 2022-11-07
Feature/mtsdk 108 access denied

crumpf:
With this PR I'm trying to commonize how the Android and iOS websockets deal with a the failure to open a connection if they supply an invalid deploymentId. Sadly, they don't respond the same way if the service forbids access.

On Android, okhttp returns a `403` error with a message "User is not authorized to access this resource with an explicit deny".

On iOS, the Apple websocket returns a error code `-1011` (`NSURLErrorBadServerResponse`) and I discovered a key in the associated `userInfo` dictionary called `_NSURLErrorWebSocketHandshakeFailureReasonKey` which I further used to constrain things.

Now, in these situations the PlatformSocket actuals will fail with a new `ErrorCode.WebsocketAccessDenied` error which the `MessagingClientImpl` can handle similarly to `ErrorCode.NetworkDisabled` and not trigger the reconnection logic and go straight to an error state.


AfanasievAnton (approved):

crumpf:
Jenkins is happy

PR #130, opened 2022-11-08
MTSDK-115: Verify Connection Closed Event

mmorehou:
New test to verify that we handle the Connection Closed Event.

Also disabling the attachment test due to it sporadically failing. Investigating and handling in a separate ticket.

### DIFF
--- a/iosApp/iosAppTests/iosAppTests.swift
+++ b/iosApp/iosAppTests/iosAppTests.swift
@@ -101,6 +101,40 @@ class iosAppTests: XCTestCase {
         contentController.disconnectMessenger()
     }
 
+    func testConnectionClosed() {
+        // Pull the deployment for use later.
+        var deployment: Deployment?
+        do {
+            deployment = try Deployment()
+        } catch {
+            XCTFail("\(error.localizedDescription)")
+        }
+        guard let deployment else {
+            XCTFail("Failed to pull the deployment config.")
+            return
+        }
+
+        // Create 4 new content controllers. We'll use these to trigger a connection closed event.
+        // Right now, there's a max number of 3 open sesssions that use the same token.
+        // We'll open 4 and confirm that an error occurs in the fourth attempt on the oldest client.
+        var controllers = [TestContentController]()
+        for _ in 1...4 {
+            controllers.append(TestContentController(deployment: deployment))
+        }
+        controllers[0].connectionClosed = XCTestExpectation(description: "Wait for the ConnectionClosedEvent")
+        for controller in controllers {
+            controller.startMessengerConnection()
+            delay(3)
+        }
+        let result = XCTWaiter().wait(for: [controllers[0].connectionClosed!], timeout: 30)
+        XCTAssertEqual(result, .completed, "Did not receive a Connection Closed event.")
+
+        // Cleanup. Also verifies that all of the other clients are still connected due to an error being thrown if we disconnect while not being connected.
+        for x in 1...3 {
+            controllers[x].disconnectMessenger()
+        }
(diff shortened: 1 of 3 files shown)


iosApp/iosAppTests/iosAppTests.swift
```diff
@@ -101,6 +101,40 @@ class iosAppTests: XCTestCase {
...
 
+    func testConnectionClosed() {
+        // Pull the deployment for use later.
+        var deployment: Deployment?
+        do {
+            deployment = try Deployment()
+        } catch {
+            XCTFail("\(error.localizedDescription)")
+        }
+        guard let deployment else {
```
crumpf, line 112: I love this new Swift feature making a shorthand for shadowing optionals that already exist, but our Jenkins agent isn't up to date yet so I fear this will cause it to complain.

mmorehou: Good point. I'll update the test to make sure an error doesn't happen.

crumpf (approved):

iosApp/iosAppTests/iosAppTests.swift
```diff
@@ -101,6 +101,40 @@ class iosAppTests: XCTestCase {
...
+        // Right now, there's a max number of 3 open sesssions that use the same token.
+        // We'll open 4 and confirm that an error occurs in the fourth attempt on the oldest client.
+        var controllers = [TestContentController]()
+        for _ in 1...4 {
+            controllers.append(TestContentController(deployment: deployment))
+        }
+        controllers[0].connectionClosed = XCTestExpectation(description: "Wait for the ConnectionClosedEvent")
+        for controller in controllers {
+            controller.startMessengerConnection()
+            delay(3)
```
AfanasievAnton, line 127: is this delay in seconds? If yes. I assume a 1 second delay is more than enough to have consistent behvaiour.

iosApp/iosAppTests/iosAppTests.swift
```diff
@@ -101,6 +101,40 @@ class iosAppTests: XCTestCase {
...
+        }
+        let result = XCTWaiter().wait(for: [controllers[0].connectionClosed!], timeout: 30)
+        XCTAssertEqual(result, .completed, "Did not receive a Connection Closed event.")
+
+        // Cleanup. Also verifies that all of the other clients are still connected due to an error being thrown if we disconnect while not being connected.
+        for x in 1...3 {
```
AfanasievAnton, line 133: Just wondering if `x` is common identifier for iterator placeholder on swift? I know in most languages the convention is to use `i`.

AfanasievAnton (approved):

iosApp/iosAppTests/iosAppTests.swift
crumpf: It might even be more conventional in Swift to be more verbose with iterator names, e.g.

```
for index in 1...3 { … }
```

as those are the examples used in the Swift language docs.

It might also be a good idea to use the array's count to construct the range to mitigate a potential out of bounds error:

```
for index in 1..<controllers.count {
    controllers[index].disconnectMessenger()
}
```

mmorehou: Missed this comment before I merged. I'll add this in on my next test update.

iosApp/iosAppTests/iosAppTests.swift
mmorehou: The 3 seconds may be excessive, but in the event of some backend slowness, a larger delay may prevent any unintended failures.

PR #136, opened 2022-11-25
MTSDK-129 Events in history.

AfanasievAnton:
Some events like `Presence`, `Quick Reply` and more can be included in history transcript along with conventional text messages. This pr is to ensure the history response contains those events.

- Allow adding event information to the history.
- Update KDoc to include events documentation.
- Update unit tests.


mmorehou (approved):

AfanasievAnton:
Successful Jenkins build.

PR #138, opened 2022-11-30
remove iOS 13 checks: transport min version is 13

crumpf:
The min supported version of the iOS Messenger Transport xcframework is 13.0, so the checks in the iosApp testbed app for earlier iOS versions can be dropped.


AfanasievAnton (approved):

bjdupuis (approved):

crumpf:
Jenkins is pleased.

PR #142, opened 2022-12-06
Exclude licence from android packaging options as junit has conflicts.

AfanasievAnton:
- Exclude licence from android packaging options as junit has conflicts.


AfanasievAnton:
Jenkins build was successful

PR #144, opened 2022-12-07
MTSDK-123: iOS Tests

mmorehou:
New iOS tests for agent name and avatar.


mmorehou:
@AfanasievAnton  I can add a check for originatingEntity as well. That shouldn't be difficult. 
I'll also look into what it would take to add a bot as well. I don't believe that's setup in the test org just yet.

AfanasievAnton (approved):

betsyknoke (approved):

PR #148, opened 2023-01-06
Feature/mtsdk 11 handle read only field from session response

AfanasievAnton:
As part of Conversation Disconnect feature the `readOnly` field was added to the ConfigureSession response on the Shyrka end. The purpose of this field is to indicate whether configured session allows further communication with an agent (`readOnly=false`) or it is allowed to only fetch conversation history to display to the client (`readOnly=true`). Note, that any attempt to send a message while `readOnly=true` will result in error response from Shyrka saying that communication is not allowed. 

- Parse `readOnly` field from SessionResponse.kt
- Expose to the client the `readOnly` field as part of State.Configure.
- Update Android and iOS Testbed application to include `readOnly` field.
- Update unit tests to include `readOnly` field.
- Provide KDoc that describes the usage of `readOnly` field.


AfanasievAnton:
Checked-out from the wrong branch. Closing this pr and will submit the same code from correct branch.

PR #149, opened 2023-01-06
MTSDK-11 Handle ReadOnly from SessionResponse.

AfanasievAnton:
As part of Conversation Disconnect feature the readOnly field was added to the ConfigureSession response on the Shyrka end. The purpose of this field is to indicate whether configured session allows further communication with an agent (readOnly=false) or it is allowed to only fetch conversation history to display to the client (readOnly=true). Note, that any attempt to send a message while readOnly=true will result in error response from Shyrka saying that communication is not allowed.
- Parse `readOnly` field from SessionResponse.kt
- Expose to the client the `readOnly` field as part of State.Configure.
- Update Android and iOS Testbed application to include `readOnly` field.
- Update unit tests to include `readOnly` field.
- Provide KDoc that describes the usage of `readOnly` field.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
@@ -37,8 +37,15 @@ interface MessagingClient {
          *
          * @property connected true if session has been configured and connection is established.
          * @property newSession indicates if configured session is new. When configuring an existing session, [newSession] will be false.
+         * @property readOnly represents the session state. When [readOnly] is set to true, the user will not be able to send or receive any new messages,
+         * but they will still be able to fetch and read existing message history.
+         * When [readOnly] is set to false, the user will be able to send and receive messages.
          */
-        data class Configured(val connected: Boolean, val newSession: Boolean) : State()
+        data class Configured(
+            val connected: Boolean,
+            val newSession: Boolean,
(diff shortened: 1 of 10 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
```diff
@@ -37,8 +37,15 @@ interface MessagingClient {
          *
          * @property connected true if session has been configured and connection is established.
          * @property newSession indicates if configured session is new. When configuring an existing session, [newSession] will be false.
+         * @property readOnly represents the session state. When [readOnly] is set to true, the user will not be able to send or receive any new messages,
+         * but they will still be able to fetch and read existing message history.
+         * When [readOnly] is set to false, the user will be able to send and receive messages.
          */
-        data class Configured(val connected: Boolean, val newSession: Boolean) : State()
+        data class Configured(
```
crumpf, line 44: I'm supposing this API change will necessitate a change our SDK's major version number?

AfanasievAnton: Not exactly. It does not modify or break any existing behaviour. Instead, it is just extend it by providing additional information. User can decide to ignore this field, therefore it does not considered to be a breaking change.

mmorehou (approved):

AfanasievAnton:
Jenkins run a successful build.

PR #150, opened 2023-01-13
MTSDK-12 Handle Disconnect event.

AfanasievAnton:
### Changes
- Add State.ReadOnly. MessagingClient will transition to this state once ConversationDisconnect event was received AND conversation disconnect is enabled in DeploymentConfig OR when SessionResponse has `readOnly=true` (Note ConversationAutostart would not be dispatched if `readOnly=true` as this is considered to be a "write/send" action. Once in this state, MessagingClient will allow read-only actions like fetch history.
- Throw InvalidStateException when application tries to invoke write/send action while MessagingClient in State.ReadOnly.
- Remove `readOnly` field from State.Configured payload. Instead MessagingClient itself will transition to State.ReadOnly once received.
- Add event ConversationDisconnect.
- Notify application with ConversationDisconnect event when it was received AND conversation disconnect is enabled in DeploymentConfig.
- Change PresenceEvent.Type from String to enum.
- Add/Update KDocs.
- Add/Update unit tests.
- Update Testbed applications on Android and iOS.


mmorehou (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/StateMachineImpl.kt
```diff
@@ -58,13 +57,21 @@ internal class StateMachineImpl(
...
+        currentState = State.ReadOnly
+    }
 }
 
 internal fun StateMachine.isConnected(): Boolean = currentState is State.Connected
 
 @Throws(IllegalStateException::class)
 internal fun StateMachine.checkIfConfigured() =
-    check(currentState is State.Configured) { "WebMessaging client is not configured." }
+    check(currentState is State.Configured) { "MessagingClient is not configured or in ReadOnly state." }
```
crumpf, line 70: maybe "configured" in the string literal here should be "Configured"

crumpf (approved):

AfanasievAnton:
Jenkins had a green build.

PR #153, opened 2023-01-19
Add tests for Participant name and avatar feature

betsyknoke:
Add test for unknown agent.
Add test for bot agent.
Update send message test to receive a return message and to verify participant name and avatar.
Add ability to enter deployment ids from config file in order to use multiple deployment ids.
Accommodate running tests in both a dev and prod region.
Add function to send an outbound message via api.


AfanasievAnton (approved):

androidComposePrototype/src/androidTest/java/com/genesys/cloud/messenger/uitest/test/ComposePrototypeUITest.kt
```diff
@@ -248,4 +292,37 @@ class ComposePrototypeUITest : BaseTests() {
         }
         bye()
     }
+
+    @Test
+    fun testUnknownAgent() {
+        apiHelper.disconnectAllConversations()
+        enterDeploymentInfo(testConfig.humanizeDisableDeploymentId)
+        DefaultTokenStore("com.genesys.cloud.messenger").store(UUID.randomUUID().toString())
```
mmorehou, line 300: Very minor suggestion. Since you're sending text before answering the conversation anyway, resetting the messenger token shouldn't be required. Keeping this is totally fine though if you prefer having a different token for every test for troubleshooting purposes.

mmorehou (approved):

androidComposePrototype/src/androidTest/java/com/genesys/cloud/messenger/uitest/test/ComposePrototypeUITest.kt
betsyknoke: I see what you mean. I will take this out.

PR #154, opened 2023-01-23
MTSDK-153 upgrade to latest Ktor version

crumpf:


transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/util/Request.kt
```diff
@@ -1,12 +1,13 @@
 package com.genesys.cloud.messenger.transport.util
 
 internal object Request {
+    const val token = "00000000-0000-0000-0000-000000000000"
```
AfanasievAnton, line 4: Good idea. Was thinking of it a couple of times, but never come to do that.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/events/UserTypingProvider.kt
```diff
@@ -6,15 +6,17 @@ import com.genesys.cloud.messenger.transport.util.Platform
...
-private const val TYPING_INDICATOR_COOL_DOWN_IN_MILLISECOND = 5000L
-
 internal class UserTypingProvider(
     private val log: Log,
     private val showUserTypingEnabled: () -> Boolean,
     private val getCurrentTimestamp: () -> Long = { Platform().epochMillis() },
 ) {
     private var lastSentUserTypingTimestamp = 0L
 
+    internal companion object {
```
AfanasievAnton, line 16: That is not really critical to me, but I think that JetBrains guides to use top level declarations for this sort of usage. 
On other hand, I understand the idea behind the using the companion object here. 
Do not have any strong preference in this context.

AfanasievAnton (approved):

PR #161, opened 2023-03-02
MTSDK-140 Obtain authcode from Okta with PKCE on Android.

AfanasievAnton:
NOTE: This pr targets MTSDK-138 branch. 

- `CodeChallenge` and `CodeChallengeMethod` are hardcoded in BuildConfig. This in order to simplify the test app flow. So that instead of generating each time new challenge and then provide utility to convert it to `CodeVerifier` (will be created as part of MTSDK-144) we will use values provided in okta.properties.
- Add `loginWithPkce` command that will be responsible to launch auth code request with CODE_CHALLENGE and CODE_CHALLENGE_METHOD provided in `okta.properties`
- Add CODE_CHALLENGE and CODE_CHALLENGE_METHOD to Jenkinsfile and env variables.
- Slightly change the way we read `deploymentId` and `deploymentDomain` in TestBedViewModel.kt init block to rely on BuildConfig if this values were not set in the selection menu. This to ease on how we set this values.
- `onOktaSignIn` function now contains the url to use for authCode request, and TestBedViewModel.kt is responsible to build and provide one based on PKCE state.
- Print authCode in `response` window for demo/qa purposes.


mmorehou (approved):

crumpf (approved):

AfanasievAnton:
Jenkins run a successful build.

PR #164, opened 2023-03-16
MTSDK-143 Configure Authenticated session.

AfanasievAnton:
NOTE: This PR is dependent on MTSDK-144 and therefore targets it. 

When `MessagingClient.configureAuthenticated` is called MessagingClientImpl will store the provided AuthJwt. After successful transition to State.Connected it will decide based on `AuthJwt` nullability if configure request should be constructed for `authenticated` or `anonymous` session. Obviously, in case `AuthJwt` is null - anonymous configure session request will be sent, otherwise `authenticated`. `AuthJwt` will be cached for purpose of proper reconnection handling, but if MessagingClient has transitioned to State.Closed or State.Error it will reset `AuthJwt` back to null. 

- Add public API to MessagingClient.kt to allow configuration of authenticated session.
- Add `configureAuthenticated` command to Android TestBedViewModel.kt
- Add ConfigureAuthenticatedSessionRequest.kt data class to build a request.
- Add error handling block when okta sign in flow result in error.
- Add KDoc to newly added public api.


mmorehou (approved):

crumpf (approved):

AfanasievAnton:
Jenkins had a successful build

PR #165, opened 2023-03-24
MTSDK-146 Logout from Authenticated Session

AfanasievAnton:
- Add `MessagingClient.logoutAuthenticated()` public API to perform logout from auth provider. Note, that actual logout request will be performed only if current session was configured as authenticated. Calling this function on anonymous session will take no effect.
- Add LogoutUseCase.kt to execute logout request.
- If MessagingClient receives a Client/Server/Redirect ResponseError during reconnection it indicates that `ConfigureRequest` has failed and therefore there are no reason to continue to Reconnect and MessagingClient should be transitioned to State.Error
- Prevent MessagingClient from handling WebSocket errors once it is in inactive state (Idle, Closing, Closed, Error)
- log any unhandled message received from Shyrka.
- Add `oktaLogout` command to Android Testbed app.
- Once logout command was successfully executed remove `authToken` from sharedprefs and indicate a state transition in UI by setting it to AuthState.LoggedOut
- Add KDoc for `MessagingClient.logoutAuthenticated()` public API.
- Add unit tests coverage.


mmorehou (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -208,6 +211,15 @@ internal class MessagingClientImpl(
             }
     }
 
+    @Throws(Exception::class)
+    override suspend fun logoutFromAuthenticatedSession() {
+        if (authJwt == null) {
+            log.w { "Logout from anonymous session is not supported." }
+            return
+        }
+        logoutUseCase.logout(authJwt!!.jwt)
```
crumpf, line 220: I see the not-null assertion is used because smart-cast isn't possible here due to `authJwt` mutability. Would it be possible to use a shadow here? e.g.:

```
    @Throws(Exception::class)
    override suspend fun logoutFromAuthenticatedSession() {
        val authJwt = authJwt ?: run {
            log.w { "Logout from anonymous session is not supported." }
            return
        }
        logoutUseCase.logout(authJwt.jwt)
    }
```

androidComposePrototype/src/main/java/com/genesys/cloud/messenger/androidcomposeprototype/ui/testbed/TestBedFragment.kt
```diff
@@ -9,12 +9,18 @@ import androidx.browser.customtabs.CustomTabsIntent
...
 import kotlinx.coroutines.runBlocking
 
 class TestBedFragment : Fragment() {
 
     private val viewModel: TestBedViewModel by activityViewModels()
     private val onOktaSignIn: (url: String) -> Unit = { url -> launchCustomTabs(url) }
+    private val onOktaLogout: () -> Unit = {
+        context?.getSharedPreferences()?.edit()?.run {
+            remove("authCode")
+        }
```
bjdupuis, line 22: I think you also need a `commit()` or `apply()` here.

bjdupuis (changes requested):

androidComposePrototype/src/main/java/com/genesys/cloud/messenger/androidcomposeprototype/ui/testbed/TestBedFragment.kt
AfanasievAnton: > I think you also need a `commit()` or `apply()` here.

Good point there. How did I even missed that one...

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
AfanasievAnton: will do

bjdupuis (approved):

PR #176, opened 2023-05-17
Feature/mtsdk 145 handle token expire

AfanasievAnton:
`AuthHandlerImpl` serves as the primary class for handling authentication-related logic and use cases.
Following the exchange of the `authCode` for a `jwt`, the `jwt` remains valid for 15 minutes by default. If a session is already configured with this jwt, the conversation can continue even after the token has expired, as the token is only necessary for session configuration.
However, if a reconnection event occurs or the client decides to close and open a new session, the `jwt` will no longer be valid. In such cases, I have implemented the refreshToken behavior. When the `authCode` is exchanged, a `refreshToken` is also received alongside the `jwt`. Unlike the `jwt`, the `refreshToken` is a long-lived token that remains valid for 24 hours and can be used to refresh the validity of the existing` jwt`. (Note that upon successful exchange, a new jwt will be received, which remains valid for 15 minutes as well.)
Once the validity of the `refreshToken` expires request to `refreshToken()` will result in error, which in turn cause the MessagingClient transition to `State.Error`, accompanied by the ErrorCode `RefreshAuthTokenFailure`, `AuthFailure`, or `AuthLogoutFailure`. The user will then have to go through the authentication process. To facilitate this behavior, the `refreshToken` is stored/updated in SharedPrefs/Keychain and retrieved when a `MessagingClient` instance is created.
In the event of a logout or failure to refresh, the Transport will remove the`refreshToken` from SharedPrefs/Keychain as it is no longer valid. In this scenario, the user will need to go through the authentication process again. Additionally, having the `refreshToken` allows the user to configure an authenticated session multiple times during the lifespan of the `refreshToken` without needing to re-authenticate. This enhances the overall user experience.
There is an option to disable the `refreshToken` behavior by setting the `Configuration.autoRefreshTokenWhenExpired` to `false`. In such cases, when a session is closed/reopened or a reconnection occurs, the user will need to re-authenticate, and the `refreshToken` will not be stored in SharedPrefs/Keychain.
 

### Auth:
- Add AuthHandler/AuthHandlerImpl to encapsulate authentication related logic.
- Handle authTokenExpire.
- Add public API `authenticate()` to MessagingClient.kt. This API will be responsible for the authCode to jwt exchange.
- Add connectAuthenticated flag to MessagingClientImpl to know when attempt connectAuthenticated
- MessagingClientImpl.configureSession for connectAuthenticated will attempt to refresh token if no jwt was set.
- MessagingClientImpl.kt -> when ClientResponseError during connection/reconnection/start of new session AND error code indicates `AuthTokenExpire` try to refresh auth token.
- Clear refresh token from both token store and local cache when refreshToken request fails.
- Throw IllegalStateException when trying to call `MessagingClient.logout` without configuring ws first.
- Add `autoRefreshTokenWhenExpired` to `Configuration` object that defines refresh flow behaviour.
- Don't refresh auth token if this behaviour is disabled in `Configuration`.
- Add Event.Authenticated. Will be sent when auth code was successfully exchanged for jwt.
- Add new ErrorCodes: `AuthFailed`, `AuthLogoutFailed` and `RefreshAuthTokenFailure` to represent failure in auth flow and `CancellationError` to ease on handling of CancellationExceptions.
- Add CorrectiveAction.Reauthenticate to request re-authentication.
- Add CorrectiveAction for ErrorCode.AuthFailed,ErrorCode.AuthLogoutFailed,ErrorCode.RefreshAuthTokenFailure
- Add functionality to store/fetch auth refresh token from/to TokenStore.
- Add `storeAuthRefreshToken` and `fetchAuthRefreshToken` in TokenStore to store and retrieve refreshToken (if enabled in AuthConfig)

### Refactor:
- Update EventHandlerImpl.kt `onEvent` api. Instead of accepting `StructuredEvent` it should accept `Event`. This for sake of simplicity.
- Cleanup the event system
- Make StructuredMessageEvent parsing eventType agnostic and resilient for the types that are not known to the system. In cases when serializer could not detect the eventType it will be deserialized as StructuredMessageEvent.Unknown and later on filtered out from the flow.
-  Use polymorphic serialization for Content with `UnknownContent` as default/fallback serializer. This in order to make serialization more resilient when new/unknown content types are introduced. 
- Add generic `Result<T>` to use for internal callback communication.
- Add class Empty to populate Result.Success when there are No Body expected in response.
- Make sure that ALL Log.Tags are starting with `MMSDK` so that it will be easier to filter logcat on Transport logs only.
- Share the same Json ContentNegotiation setup for all network clients.
- Add public api documentation.

## TestbedApp
- Instead of storing authCode in SP, set it directly to the VM. It is anyway a short leaving code and there are no good reason to store it.
- Move the TestBedVM.init to the onCreate of TestBedFragment. Having it on resume mess up the states and reinit a new MessagingClient isntance for no good reason.
- Use WeakReference when store Context to avoid potential memory leak.
- Refactor TestBedViewModel. Remove `onOktaLogout`. No need to rely on authJwt to perform auth operations as it will be cached and handled inside Transport. Handle 
- Remove onResume from TestBedFragment.


mmorehou (approved):

bjdupuis (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/auth/AuthHandler.kt
```diff
@@ -0,0 +1,17 @@
+package com.genesys.cloud.messenger.transport.auth
+
+import com.genesys.cloud.messenger.transport.core.Empty
+import com.genesys.cloud.messenger.transport.core.Result
+
+internal const val NO_JWT = "ZW1wdHlfand0"
```
crumpf, line 6: I was wondering, what are these values for `NO_JWT` and `NO_REFRESH_TOKEN`?

AfanasievAnton: These values represents the state of jwt and refreshToken when they are either unknown, unset or when there were a critical failure related to auth flow. Potentially I could use null, but I wanted to avoid null checks.

crumpf: I was concerned the string values had a special meaning, but after we talked I see that they're just base64 encodings approximating the variable names 😀

PR #177, opened 2023-05-29
Feature/mtsdk 139 obtain auth code from okta i os

AfanasievAnton:
- Implement both `oktaSignIn` and `oktaSignInWithPKCE`.
- Implement auth state view in ios demo app.
- Setup a deeplink url.
- When `authCode` received - update TestbedViewController.swift with the value.
- Perform `disconnect` when LogoutEvent received.
- Add Okta.plist to store all auth variables required for auth flow.
- Add commands for authenticated actions.
- Bridge between auth api and ios app calls.
- Slightly improve logs when configureSession() is called.
- Add Event.Logout handling.


mmorehou (approved):
Not seeing any issues when running through existing tests and from manually going through authentication. New tests are coming in a separate PR.

iosApp/Podfile
```diff
@@ -1,3 +1,4 @@
+source 'https://cdn.cocoapods.org/'
```
crumpf, line 1: I don't think it should be necessary to `source` the CocoaPods CDN. The official CocoaPods source is implicit unless another source is also used, in which case it must then be added as well.

iosApp/iosApp/SceneDelegate.swift
```diff
@@ -45,7 +45,24 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         // Use this method to save data, release shared resources, and store enough scene-specific state information
         // to restore the scene back to its current state.
     }
-
-
+    
+    func scene(_ scene: UIScene, openURLContexts URLContexts: Set<UIOpenURLContext>) {
```
crumpf, line 49: Thinking it might be possible refactor some of the nested `if` conditionals here. I think the following is logically consistent but coalesces some of the `if`'s to reduce the nesting.

```
    func scene(_ scene: UIScene, openURLContexts URLContexts: Set<UIOpenURLContext>) {
        guard let urlContext = URLContexts.first else {
            return
        }

        if let components = URLComponents(url: urlContext.url, resolvingAgainstBaseURL: false),
           let queryItems = components.queryItems {
            for item in queryItems {
                if item.name == "code", let code = item.value {
                    if let windowScene = scene as? UIWindowScene,
                       let rootViewController = windowScene.windows.first?.rootViewController as? TestbedViewController {
                        rootViewController.setAuthCode(code)
                    }
                }
                break
            }
        }
    }
```

iosApp/iosApp/TestbedViewController.swift
```diff
@@ -197,19 +224,33 @@ class TestbedViewController: UIViewController {
         case let typing as Event.AgentTyping:
             displayEvent = "Event received: \(typing.description)"
         case let error as Event.Error:
+            if(error.errorCode is ErrorCode.AuthFailed
```
crumpf, line 227: in Swift, parens enclosing the `if` condition are not necessary

iosApp/iosApp/TestbedViewController.swift
```diff
@@ -269,7 +310,48 @@ class TestbedViewController: UIViewController {
...
+                let plistDictionary = try? PropertyListSerialization.propertyList(from: plistData, options: [], format: nil) as? [String: Any],
+                let oktaDomain = plistDictionary["oktaDomain"] as? String,
+                let clientId = plistDictionary["clientId"] as? String,
+                let signInRedirectURI = plistDictionary["signInRedirectURI"] as? String,
+                let scope = plistDictionary["scopes"] as? String,
+                let oktaState = plistDictionary["oktaState"] as? String,
+                let codeChallengeMethod = plistDictionary["codeChallengeMethod"] as? String,
+                let codeChallenge = plistDictionary["codeChallenge"] as? String
+        else {
+            fatalError("Unable to read Okta.plist or missing required key")
```
crumpf, line 326: I see `fatalError` is being used here and below. `fatalError` will stop execution of the app too. Is that the intention, or is there a way we can report the error and not have to halt the app.

AfanasievAnton: You are right. I will print this error to the screen instead.

iosApp/iosApp/TestbedViewController.swift
```diff
@@ -138,8 +155,23 @@ class TestbedViewController: UIViewController {
         messenger.eventSubject
             .sink(receiveValue: updateWithEvent)
             .store(in: &cancellables)
+            authStateView.text = "Auth State: \(authState)"
     }
-    
+
+    private func signIn() {
```
crumpf, line 161: I think you could also change `buildOktaAuthorizeUrl() -> String?` to `buildOktaAuthorizeUrl() -> URL?` (doing the `URL(string: authUrlString!)` URL creating inside the builder func) and then simplify this `signIn` func logic to:

```
    private func signIn() {
        guard let authUrl = buildOktaAuthorizeUrl() else {
            authState = AuthState.error(errorCode: ErrorCode.AuthFailed.shared, message: "Failed to build Okta authorize URL.", correctiveAction: CorrectiveAction.ReAuthenticate.shared)
            updateAuthStateView()
            return
        }
        
        if UIApplication.shared.canOpenURL(authUrl) {
            UIApplication.shared.open(authUrl)
        }
    }
```

AfanasievAnton: Done

crumpf (approved):

PR #183, opened 2023-06-09
MTSDK-187 duplicated symbols when transport used in intermediary framework

crumpf:
Patches an issue introduced in v2.3.0 where the CocoapodsExtension was configured to be static. This caused duplicated symbols when transport was named as a dependency in an intermediary framework and also depended on by an enclosing app using both transport and the intermediary frameworks.

* configure CocoapodsExtension to `isStatic = false`
* bump version to 2.3.2

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -26,7 +26,7 @@ plugins {
 }
 
 // CocoaPods requires the podspec to have a `version`.
-val buildVersion = "2.3.1"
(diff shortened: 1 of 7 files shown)


AfanasievAnton (approved):

bjdupuis (approved):

mmorehou (approved):

crumpf:
Jenkins had a successful run.

PR #186, opened 2023-06-13
use BuildKonfig to provide sdk build configuration details

crumpf:
Sometimes I was finding that I'd like to be able to fetch some build config details at runtime. This PR hopes to provide some structure for being able to include that kind of information in the project. This might be able to support (or resolve?) MTSDK-53 too.


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessengerTransport.kt
```diff
@@ -28,6 +29,16 @@ class MessengerTransport(
 
     constructor(configuration: Configuration) : this(configuration, DefaultTokenStore(TOKEN_STORE_KEY))
 
+    /**
+     * The product name.
+     */
+    val name = BuildKonfig.name
```
AfanasievAnton, line 35: I was thinking, maybe we should extract this values to a separate object that can be accessed statically, without a need to create an instance of MessengerTransport for that?

crumpf: Would it make sense to do this as a companion object on `MessengerTransport` instead of making a separate object?

PR #193, opened 2023-07-11
MTSDK-200 Do not check DeploymentConfig.

AfanasievAnton:
- Remove checks of DeploymentConfig.ConversationDisconnect status when PresenceEvent.Disconnect arrives.
- When `readOnly` is not present in PresenceEvent.Disconnect metadata, then there are no need to fallback to verifying DeploymentConfig.ConversationDisconnect status.
- Remove `isConversationDisconnectEnabled`and `isReadOnly` utility functions, as there is no need for them.
- Remove ir-relevant unit tests.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
@@ -18,7 +18,6 @@ import com.genesys.cloud.messenger.transport.network.WebMessagingApi
 import com.genesys.cloud.messenger.transport.shyrka.WebMessagingJson
 import com.genesys.cloud.messenger.transport.shyrka.receive.AttachmentDeletedResponse
 import com.genesys.cloud.messenger.transport.shyrka.receive.ConnectionClosedEvent
-import com.genesys.cloud.messenger.transport.shyrka.receive.Conversations
 import com.genesys.cloud.messenger.transport.shyrka.receive.DeploymentConfig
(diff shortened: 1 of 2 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -386,15 +385,8 @@ internal class MessagingClientImpl(
                 if (structuredMessage.isOutbound()) {
                     structuredMessage.events.forEach {
                         if (it.isDisconnectionEvent()) {
-                            if (deploymentConfig.isConversationDisconnectEnabled()) {
```
tehilaR2, line 389: Why this is not needed anymore?
The BE forces session status (like readOnly) no matter if the feature is enabled on the console or if we got it, it must be enabled.

AfanasievAnton: For the Conversation.Disconnect flow we should treat BE (Shyrka) as an only source of truth. DeploymentConfig might be changed at any point, but we still have to respect whatever Shyrka sends us. 
For example: 
1. Admin has disabled ConversationDisconnect feature. 
2. Transport has fetched latest DeploymentConfig.
3. Shyrka still treat this session like ConversationDisconnect was enabled.
4. Shyrka sends us Event.ConversationDisconnect with/without readonly.
5. TransportSDK ignore this event/session state change.
6. From shyrka perspective session should be disconnected, but from Transport it is still active. 

By removing this checks we respect whatever Shyrka sends us without checking/falling back to DeploymentConfig.

tehilaR2 (approved):

PR #199, opened 2023-07-17
MTSDK-212 Send PresenceEvent.Clear

AfanasievAnton:
- Add new API: MessagingClient.clearConversation(). This function will dispatch PresenceEvent.Clear only if MessagingClient is Configured or ReadOnly and DeploymentConfig has ClearConversation enabled, otherwise fails either with exception or Error event.
- Add new PresenceEvent.Type.Clear to indicate the ClearConversation event.
- Do not dispatch any event upon receiving 200 confirmation on StructuredMessageEvent.Presence.Clear as it should be done when SessionClearedEvent is dispatched by Shyrka (will be accomplished as part of MTSDK-213)
- Add ClearConversationRequest.kt to use for sending clearConversation.
- Add ErrorCode.ClearConversationFailure to indicate ClearConversation failure.
- Add ErrorMessage.FailedToClearConversation.
- Add Unit tests to support code changes.


transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImplTest.kt
```diff
@@ -1457,6 +1457,93 @@ class MessagingClientImplTest {
...
+            connectSequence()
+            mockEventHandler.onEvent(eq(expectedEvent))
+        }
+        verify(exactly = 0) {
+            mockPlatformSocket.sendMessage(eq(Request.clearConversation))
+        }
+    }
+
+    @Test
+    fun whenSessionIsConfiguredAndClearConversationCalled() {
```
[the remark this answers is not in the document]
AfanasievAnton: Short answer - Yes.
Long one:
We have a FakeDeploymentConfig which is used for testing purposes. It has some default values, as well structure that allows overriding of this default values. 
We use default values where possible and override them only for test cases that requires different set up. Like in `whenClearConversationIsDisabledInDeploymentConfigAndClearConversationIsCalled` test case, where we want to set deploymentConfig.clearConversation.enabled to false.

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImplTest.kt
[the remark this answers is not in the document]
tehilaR2: you can have a package for major class tests as in case of the MessagingClient and have the tests under that package.
It is hard to follow all the tests in one place.
But that is my opinion

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImplTest.kt
```diff
@@ -1457,6 +1457,93 @@ class MessagingClientImplTest {
         }
     }
 
+    @Test
+    fun whenClearConversationFromAnyNonReadOnlyOrConfiguredState() {
```
tehilaR2, line 1461: What about testing the `clearConversation` activation when the session is an authenticated session?
According to the feature charter, clearing an authenticated session should not be enabled.
> see Authenticated Use-cases on
 https://intranet.genesys.com/display/PCINIT/PURE-4519+-+Messenger+-+End-User+can+Clear+Guest+Session

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -596,3 +611,6 @@ private fun KProperty0<DeploymentConfig?>.isAutostartEnabled(): Boolean =
 
 private fun KProperty0<DeploymentConfig?>.isShowUserTypingEnabled(): Boolean =
     this.get()?.messenger?.apps?.conversations?.showUserTypingIndicator == true
+
+private fun KProperty0<DeploymentConfig?>.isClearConversationEnabled(): Boolean =
+    this.get()?.messenger?.apps?.conversations?.conversationClear?.enabled == true
```
tehilaR2, line 616: according to the charter this feature is disabled by default

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImplTest.kt
```diff
@@ -1457,6 +1457,93 @@ class MessagingClientImplTest {
...
+
+        assertThat(subject.currentState).isConfigured(connected = true, newSession = true)
+        verifySequence {
+            connectSequence()
+            mockPlatformSocket.sendMessage(eq(Request.clearConversation))
+        }
+    }
+
+    @Test
+    fun whenSessionIsReadOnlyAndClearConversationCalled() {
```
tehilaR2, line 1532: What is the point of clearing history of a readonly session. This make the whole session existence redundant. 
Won't it be better just to expire the session or something?

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImplTest.kt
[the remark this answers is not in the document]
AfanasievAnton: I have created a separate ticket to handle this refactor: 
https://inindca.atlassian.net/browse/MTSDK-221

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImplTest.kt
AfanasievAnton: Took me a while, but you are looking at the wrong charter. Here is the right one:
https://intranet.genesys.com/pages/viewpage.action?pageId=277722827

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
AfanasievAnton: Again. Wrong charter.

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImplTest.kt
AfanasievAnton: No matter state(as soon as it active) for privacy reasons, user must have capability to clear the conversation history.
As part of the flow SessionClearedEvent will be dispatched by Shyrka following by ConnectionClosedEvent. 
At ConnectionClosedEvent Transport will close the ws connection, forcing UI to establish new connection. 
N

AfanasievAnton: No matter state(as soon as it is active) user must have capability to clear the conversation history. This is done for privacy reasons.
Actually, once clearconversation is called some sort of session expiry does happen. As part of the flow SessionClearedEvent will be dispatched by Shyrka following by ConnectionClosedEvent.
At ConnectionClosedEvent Transport will close the ws connection, forcing UI to establish new one.
So if you do that from ReadOnly state the new configureSession request will result in new session in state Configured with all what it implies(like Autorsart request and etc).

tehilaR2:
> I would consider that as an extreme edge case and delegate the decision to timeout to the sdk consumer. In any case 200 response should not be an indicator of success for PresenceEvent.Clear

In case the clear fails on the BE, will a dedicated error be sent to the Transport SDK?
(though it's not cover the above mentioned scenario were no acknowledge of clear status is provided by the BE.

AfanasievAnton:
> 
There is no dedicated error in case of failure.

AfanasievAnton:
> In case the clear failed, there is no way to know that? So if the client activates the clear operation and is on wait status on the UI, he will be stack if the clear fails? So, timer mechanism should wait only for success or client state changes to Error. Is this missing error indication also on the conversations about BE error exposing with codes?

When a 'Clear' request fails, we'll receive a 4xx/5xx response, and we capable of handling that. However, if the request returns a successful 200 response, there's no reason why the SessionClearedEvent wouldn't be received.
The reason we need the SessionClearedEvent is because the same session can be shared across multiple devices simultaneously. So, when a User clears the conversation, all of these devices should receive this Event (only the one that sent the request will get the 200 response). I would say that if we get 200 OK but did not get SessionClearedEvent, there are some major network issue that will cause the ws to collapse.

tehilaR2 (approved):

PR #205, opened 2023-07-31
MTSDK-189 Add `application` parameter to ws request.

AfanasievAnton:
For analytics purposes new `application` parameter has been added to WS request.

- Implement new `application` query parameter to ws connection request in the format: `application=OS_NAME+TransportSDK-VERSION_NUMBER`
- Update unit test.


transport/src/androidMain/kotlin/com/genesys/cloud/messenger/transport/util/Platform.kt
```diff
@@ -11,6 +11,11 @@ internal actual class Platform {
      */
     actual val platform: String = "Android ${android.os.Build.VERSION.SDK_INT}"
 
+    /**
+     * The name of operating system running on this device.
+     */
+    actual val osName: String = "Android"
```
tehilaR2, line 17: why not using the platform property which you already have?

PR #212, opened 2023-09-01
Tech debt/mtsdk 246 send c.attrbutes on autostart

AfanasievAnton:
The main purpose of this pr is to allow sending of customAttributes along with AutostartRequest (when it is enabled in Admin Configurations). 

Added new public api to MessagingClient - CustomAttributeStore. Now Developer can use this object to set and get customAttributes. Internally we have a CustomAttributesImpl that is responsible to manage provided custom attributes so that every new message or autostart event will send new attributes (of course if there were any modifications to them)

- Add CustomAttributesStore.kt as public API to allow developers direct work with custom attributes. 
- sendMessage API will still allow passing customAttributes as optional parameter, however internally they will still go through the CustomAttributesStoreImpl.kt
- Add CustomAttributesStoreImpl.kt as internal implementation that manage custom attribute dispatch.
- Add Channel.Metadata to EventMessage.kt
- Add Channel to AutoStartRequest.kt
- Add new CorrectiveAction.CustomAttributeSizeTooLarge
- MessagingClientImpl.kt will hold `sendingAutostart` flag so that error handling of ErrorCode.CustomAttributeSizeTooLarge can be separated between regular OnMessageRequest.kt and AutoStartRequest.kt
- In case, when conversationDisconnect flow happens and `newChat` is called - TransportSDK will sent Autostart with customAttributes that are up to date.
- CustomAttributes will be cleared if they became too large. This in order to avoid situation where every sendMessage attempt fails due to custom attributes is to large.
- Change BaseMessageProtocol scope to `internal`
- Add/Update UnitTests to reflect new changes.
- Add KDoc.
- Modify Android and iOS Testbed applications to accommodate for customAttributes api changes.


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -328,6 +337,19 @@ internal class MessagingClientImpl(
             is ErrorCode.RequestRateTooHigh,
             is ErrorCode.CustomAttributeSizeTooLarge,
             -> {
+                if (code is ErrorCode.CustomAttributeSizeTooLarge) {
+                    messageStore.clearInitialCustomAttributes()
+                    if (sendingAutostart) {
+                        sendingAutostart = false
+                        eventHandler.onEvent(
```
tehilaR2, line 344: In case of `autostart` message failure due to CA, you've mentioned that the `autostart` message will be re-sent.
How is it done? The eventListener is notified of the customAttributes error, in this case. On what error channel are you sending CA errors, is it both on messageListener and eventListener? Will the same error be caught twice if listening on both listeners?

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessageStore.kt
```diff
@@ -117,6 +121,12 @@ internal class MessageStore(
             }
         }
     }
+
+    private fun mergeInitialCustomAttributesWith(customAttributes: Map<String, String>) {
```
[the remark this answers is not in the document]
AfanasievAnton: You have touched a couple of important points here, so I will. try to answer them accordingly. 


1. To remove custom attribute you need to send it with value set to empty string. We do not accept nullable CA values. 
2. The idea behind keeping initial CA and updating them with new values (where new values actually override old ones) is to handle the `newChat` scenario in a proper manner. 

Given this are initial CA that were set in `createMessagingClient`: 
mapOf("name" to "John", "location" to "USA", "foodPreference" to "Salad")
Autostart request will contain all of this data.

if later on developer decided to update "location" and remove the "foodPreference", this will be the way to do that.
sendMessage("text message here", mapOf( "location" to "Canada", "foodPreference" to ""))

so send message request will have this CA:

Channel(Metadata(customAttributes = mapOf( "location" to "Canada", "foodPreference" to "")))
Note we would not send the "name" because it was not set in sendMessage. 
However the initialCustomAttributes map will look like following:

mapOf("name" to "John", "location" to "Canada", "foodPreference" to "")

Now the ConversationDisconnect feature comes into the picture.
So if session was ended by the Agent and ConversationDisconnect is set to ReadOnly, there is an option to start  a new chat using TransportSDK api `startNewChat`. 
This api will ensure that all devices that share the same token will close existing session. And entirely fresh  new session will be configured on the device that called `startNewChat`.
If Autostart is configured to be enabled Autostart request will be dispatched as well. So if we would not update the initialCustomAttributes with values that were send later, this autostart will contain data that was set initially:
 mapOf("name" to "John", "location" to "USA", "foodPreference" to "Salad")
 The idea behind caching and updating initialCustomAttributes is to send an updated CA along with autostart. Something that looks like that:
 mapOf("name" to "John", "location" to "Canada", "foodPreference" to "")


I feel like we might want to have a short call to discuss the details of this decision.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
AfanasievAnton: When CA exceed allowed size (that is basically the only reason why CA can fail) they fail the entire request. 

In case when CA attached to Message, they will fail the message request and this error will be communicated as error from within the messageListener. 

In case when CA attached to Autostart they will fail the Autostart request and the error will be comunicated as error on eventListener. 

It is not going to be duplicated on both channels. Only one of them will be reported.

There is an existing ticket on Shyrka to improve the overall UX. I have requested to add CA size limitation to SessionResponse. So if we know the CA limit ahead of sending the request and we find out that CA exceed this limitation instead of sending the request with them (and as a result failure the whole request) I will send an error event (via eventListener) stating that CA exceed max size and strip them down from the request (either Autostart or OnMessage). In that case request will be dispatched without CA, but developer also will be notified that they do not match the size requirements. 

here is a ticket: https://inindca.atlassian.net/browse/SHYRKA-2182

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessageStore.kt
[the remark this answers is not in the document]
tehilaR2: yes, I guess we should. I don't understand why you don't just use the same set of attributes and updates it according to the sendMessage CA if given.

AfanasievAnton:
@tehilaR2 @KalmanPerkelGenesys PR is updated according to what we discussed. Took a while to refactor the unit tests...

KalmanPerkelGenesys (approved):

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/messagingclient/MCMessageTests.kt
```diff
@@ -65,6 +71,9 @@ class MCMessageTests : BaseMessagingClientTest() {
             connectSequence()
             mockMessageStore.onMessageError(ErrorCode.MessageTooLong, "message too long")
         }
+        verify(exactly = 0) {
+            mockCustomAttributesStore.onError()
```
tehilaR2, line 75: should you add a verification for the CA state? it should be pending or sending?

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/messagingclient/MCMessageTests.kt
```diff
@@ -84,5 +93,32 @@ class MCMessageTests : BaseMessagingClientTest() {
...
+        slot.captured.onMessage(Response.customAttributeSizeTooLarge)
+
+        verifySequence {
+            connectSequence()
+            mockCustomAttributesStore.onError()
+            mockMessageStore.onMessageError(
+                ErrorCode.CustomAttributeSizeTooLarge,
+                expectedErrorMessage
+            )
+            mockAttachmentHandler.onMessageError(
```
tehilaR2, line 115: so in case of CA failure on message sending, the message should be resent by the SDK consumer?
You are not resending the message without the CA, if failed ?

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/messagingclient/MCMessageTests.kt
AfanasievAnton: CustomAttributesStore is not a subject for this test and is mocked. So it is not going to get the proper state transition unless I explicitly implement mock behaviour for it. Something like:

every { mockCustomAttributesStore.onError() } answers mockCustomAttributes.state = State.Error. 

Or instead of mocking, I will need to provide an actual instance of CustomAttributesStore and then test the values. 

But Since we have a dedicated CustomAttributesStoreTest that knows how to test this state transitions, I have no need to do it here.

transport/src/androidTest/kotlin/com/genesys/cloud/messenger/transport/core/messagingclient/MCMessageTests.kt
AfanasievAnton: No, because potentially, there can be multiple reasons for the failure.

tehilaR2 (approved):

PR #217, opened 2023-09-21
MTSDK-253 Expose SCP data as API

AfanasievAnton:
- Add FileAttachmentProfile.kt to represent the SCP data in more convenient way.
- Add nullable fileAttachmentProfile field to MessagingClient.kt API.
- Map AllowedMedia to FileAttachmentProfile.kt when received in SessionResponse.
- Add new Const file to hold commonly used constants.
- Add KDoc for public api
- Add unit tests to cover possible parsing/mapping scenarios.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
@@ -44,6 +44,7 @@ import com.genesys.cloud.messenger.transport.shyrka.send.JourneyCustomer
 import com.genesys.cloud.messenger.transport.shyrka.send.JourneyCustomerSession
 import com.genesys.cloud.messenger.transport.util.Platform
 import com.genesys.cloud.messenger.transport.util.Vault
+import com.genesys.cloud.messenger.transport.util.extensions.toFileAttachmentProfile
 import com.genesys.cloud.messenger.transport.util.extensions.toMessage
 import com.genesys.cloud.messenger.transport.util.extensions.toMessageList
 import com.genesys.cloud.messenger.transport.util.logs.Log
@@ -84,6 +85,7 @@ internal class MessagingClientImpl(
     private var connectAuthenticated = false
     private var isStartingANewSession = false
     private var reconfigureAttempts = 0
+    private var _fileAttachmentProfile: FileAttachmentProfile? = null
 
     override val currentState: State
         get() {
@@ -113,6 +115,9 @@ internal class MessagingClientImpl(
 
     override val conversation: List<Message>
         get() = messageStore.getConversation()
+
+    override val fileAttachmentProfile: FileAttachmentProfile?
+        get() = _fileAttachmentProfile
 
     @Throws(IllegalStateException::class)
     override fun connect() {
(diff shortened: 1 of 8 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -114,6 +116,9 @@ internal class MessagingClientImpl(
     override val conversation: List<Message>
         get() = messageStore.getConversation()
 
+    override val fileAttachmentProfile: FileAttachmentProfile?
```
KalmanPerkelGenesys, line 119: This is an elegant way of exposing this information. However, how will the integrating developer know when this information is ready and not null? The only option available to them is to check the documentation and see that it is available when they receive `MessagingClient.State.Configured`.

Additionally, there are edge cases to handle, such as reconnection and session expiration. In all of these cases, the developer will need to check if there were changes in the `fileAttachmentProfile`

Wouldn't it be better to handle this information in a similar to how we handle other 'SessionResponse' data? We could pass 'fileAttachmentProfile' or combine all the info into another formatted data (like 'sessionData') through 'stateMachine.onSessionConfigured(connected, newSession).

AfanasievAnton: Thanks for the feedback!
Passing FileAttachmentProfile with `State.Configure` was one of the options I have been considering. 
But there were a couple of downsides for this approach.

1. Developer will have to cache the FileAttachmentProfile on his/her side and manage/refresh it once new data has arrived, which seems a little bit redundant and error prompt to me. In current approach there is no need in doing that. And that includes reconnection and session expiry. Having this data exposed as field, guarantee that SCP is always up-to-date when you need it. Similar to `sendMessage` or `attach` - we do not expose this functionality upon session Configure, but we still prohibit the usage of this api's when MessagingClient state does not allow that.

2. Except of `State.Configured` there is also another `State.ReadOnly` that can provide/expose SCP data. It is true, that while in this state developer would not be able to attach anything anyway, but this is just an example of a potential use-case. We have no guarantee that in the future we would not introduce new states that potentially could have SCP data in them, and then it can cause confusion. 

3. There is another BE api `GetConfigurations` that provides additional way to fetch the SCP data. At this time TransportSDK has no plans to implement this API because there is no valid use-case for us. But if we decide that at some point we do that(for any reason) it will be much more complicated to expose FileAttachmentProfile to the developer. As with proposed solution it will be sticked to Configure state. So we will have to introduce some sort of new state and/or event to notify developer that SCP was updated/refreshed, while with proposed approach it will be performed seamlessly (the existing fileAttachmentProfile data will be replaced with new one).

I can add another kdoc line to `MesssagingClient.fileAttachmentProfile : FIleAttachmentProfile?` to explain that this value is lateinit and will be available for access only once MessagingClient is configured.

KalmanPerkelGenesys: Thank you for the clear answers

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -84,6 +85,7 @@ internal class MessagingClientImpl(
     private var connectAuthenticated = false
     private var isStartingANewSession = false
     private var reconfigureAttempts = 0
+    private var _fileAttachmentProfile: FileAttachmentProfile? = null
```
KalmanPerkelGenesys, line 88: Why is there a need to have a private `_fileAttachmentProfile` and a public `fileAttachmentProfile` variables?

KalmanPerkelGenesys (approved):

PR #219, opened 2023-09-28
MTSDK-238 Validate file size before upload

AfanasievAnton:
- Add `validate(ByteArray)` function to AttachmentHandler/Impl to verify that provided ByteArray size match `maxFileSizeKB`
- Move reference to FileAttachmentProfile from MessagingClientImpl.kt to AttachmentHandlerImpl.kt as it makes more sense to have it there.
- Add `throw IllegalArgumentException` to `attach` function and throw it when `attachmentHandler.validate()` returns false.
- Add dedicated ErrorMessage when file size exceed max allowed size.
- Update existing unit tests to accommodate to the changes.
- Update some existing unit test names to use more convenient "back-tick" kotlin feature for naming functions.
- Update KDoc to include the new throw type of attach() function

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/AttachmentHandlerImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/AttachmentHandlerImpl.kt
@@ -13,10 +13,8 @@ import com.genesys.cloud.messenger.transport.shyrka.receive.UploadSuccessEvent
 import com.genesys.cloud.messenger.transport.shyrka.send.DeleteAttachmentRequest
 import com.genesys.cloud.messenger.transport.shyrka.send.OnAttachmentRequest
 import com.genesys.cloud.messenger.transport.util.logs.Log
-import io.ktor.client.plugins.ResponseException
 import io.ktor.http.ContentType
 import io.ktor.http.defaultForFilePath
-import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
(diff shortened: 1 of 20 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/AttachmentHandlerImpl.kt
```diff
@@ -106,6 +108,12 @@ internal class AttachmentHandlerImpl(
         return null
     }
 
+    override fun validate(byteArray: ByteArray): Boolean {
```
KalmanPerkelGenesys, line 111: Since this function is for validating the size of the file attachment, I think it would be clearer to name it `validateSize` or something similar

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -190,12 +189,16 @@ internal class MessagingClientImpl(
...
     }
 
+    @Throws(IllegalStateException::class, IllegalArgumentException::class)
     override fun attach(
         byteArray: ByteArray,
         fileName: String,
         uploadProgress: ((Float) -> Unit)?,
     ): String {
         log.i { "attach(fileName = $fileName)" }
+        if (!attachmentHandler.validate(byteArray)) {
```
KalmanPerkelGenesys, line 199: I don't think it's bad that this code is here, but I'm wondering if it can be moved to the `prepare()` function. Do you need a public `validate()` function for use in other places?

Maybe moving this into the attachment handler will make the file preparation process more encapsulated and minimize the amount of code in the Messaging client.

KalmanPerkelGenesys (approved):
Please see my suggestions

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/AttachmentHandlerImpl.kt
AfanasievAnton: filetype and blocked extension validation will be added to the same function, so the name will remain the same, but validation condition will be extended.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
AfanasievAnton: hmmm...totally make sense! Will move it to `prepare`

PR #220, opened 2023-09-29
MTSDK-245 Implement refreshAttachmentUrl api

AfanasievAnton:
NOTE: MTSDK-245 was checked out from MTSDK-238 (the validate filesize) therefore I am merging it back. 
In perfect case when MTSDK-238 will be merged into main feature branch (MTSDK-137) I will rebase/update current branch and change its target to main feature branch. 

PS. Sorry for confusion.

- Implement refresh downloadUrl flow.
- Add new `refreshAttachmentUrl()` api to MessagingClient.kt
- Add new Attachment.State.Refreshed. Attachment will transition to this state upon successful refresh of downloadUrl and UI will be notified.
- Refactor `WebMessagingApi.uploadFile()` to align with structure of other REST api requests.
- Add/Update unit tests.
- Add KDoc for `refreshAttachmentUrl` and `Attachment.State.Refreshed`.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
@@ -196,6 +196,16 @@ interface MessagingClient {
     fun detach(attachmentId: String)
 
     /**
+     * Refresh the downloadUrl.
+     *
+     * @param attachmentId the ID of the attachment to refresh.
+     *
+     * @throws IllegalStateException If the current state of the MessagingClient is not compatible with the requested action.
+     */
+    @Throws(IllegalStateException::class)
+    fun refreshAttachmentUrl(attachmentId: String)
+
+    /**
      * Get message history for a conversation.
      *
      * @throws Exception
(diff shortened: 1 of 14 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
```diff
@@ -195,6 +195,16 @@ interface MessagingClient {
     @Throws(IllegalStateException::class)
     fun detach(attachmentId: String)
 
+    /**
+     * Refresh the downloadUrl.
```
KalmanPerkelGenesys, line 199: Do we need to add here explanation of why we need this? I mean, an explanation about when the URL expires and such

KalmanPerkelGenesys (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
AfanasievAnton: It is a good point. I will add a sentence about it .

PR #221, opened 2023-10-04
MTSDK-242 Validate blocked extensions.

AfanasievAnton:
- Add blocked extension validation on the file name
- Slightly refactor `AttachmentHandlerImpl.validate()` function to improve readability.
- Add a couple of extension function to help with file name and size validation.
- Validate that attachment size is not 0 and provide a meaningful error message if so.
- Add/Update unit tests.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/AttachmentHandlerImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/AttachmentHandlerImpl.kt
@@ -39,9 +39,7 @@ internal class AttachmentHandlerImpl(
         fileName: String,
         uploadProgress: ((Float) -> Unit)?,
     ): OnAttachmentRequest {
-        if (!validate(byteArray)) {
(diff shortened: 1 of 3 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/AttachmentHandlerImpl.kt
```diff
@@ -177,6 +186,12 @@ internal class AttachmentHandlerImpl(
     }
 }
 
+private fun ByteArray.isInvalid(maxFileSizeKB: Long?): Boolean =
```
KalmanPerkelGenesys, line 189: Why do you check if it is invalid rather than checking if it is valid?

AfanasievAnton: Great observation.

The choice to check if the byteArray is invalid rather than valid was influenced by the usage context. Since a failure to validate the file size triggers an exception, the condition reads more intuitively as: "if byteArray is invalid, then throw." It adds a level of clarity to the logic, making it easier to follow (at least on my opinion).
 I find it more readable than using !valid.

PR #224, opened 2023-10-10
MTSDK-237 Parse new fields from DeploymentConfig

AfanasievAnton:
- Add parsing of `enableAttachments` field in DeploymentConfig with fallback to `false` if not available or failed to parse.


AfanasievAnton:
> In the JIRA issue I ca see that the fileTypes inside Mode is optional, it is not present in the second scenario. This is correct? In this case the fileTypes in line 79 shouldn't be optional?

Good point. I will add a default fallback to empty list in case it is not present in the response.

monitamasi (approved):

PR #225, opened 2023-10-17
Feature/mtsdk 241 implement scp on i os testbed

AfanasievAnton:
- Implement file picker on iOS Testbed app.
- Implement refreshAttachment url on iOS Testbed.
- Remove `selectAttachment` command as it is redundant now.


mmorehou (approved):
Verified that existing tests including attachment related tests are passing. Manually verified that Source Content Profiles are being pulled and that we're handling errors when we attempt to send an invalid attachment type.

monitamasi (approved):

PR #253, opened 2024-01-30
Feature/mtsdk 198 implement quick replies on i os testbed app

AfanasievAnton:
- Store quick replies as map when received.
- Implement sendQuickReply command.
- Add sendQuickReply command to MessengerInteractor.swift
- Add print of Event.ConversationDisconnect that was missing on iOS.
- Align print of quick replies received message between iOS and Android.


mmorehou (approved):
Verified that quick replies appear to be working in the iOS testbed app. Nice work!

ksw5 (approved):

PR #256, opened 2024-02-01
Adding test for Quick Replies

betsyknoke:
New automated test for Quick Replies
Verifies response from sending a quick reply command to Carousel.
Verifies sending an invalid quick reply command returns an error.


AfanasievAnton (approved):

mmorehou (approved):
Nice work!

PR #262, opened 2024-02-28
Add additional prod regions 

betsyknoke:
Add additional prod regions to the list
Update tests to run in additional region


build.gradle.kts
```diff
@@ -31,14 +31,4 @@ val buildVersion = "2.6.0-rc2"
 val snapshot = System.getenv("SNAPSHOT_BUILD") ?: ""
 version = "${buildVersion}${snapshot}"
 group = "cloud.genesys"
-
```
AfanasievAnton, line 34: Is commenting out nexusPublishing intentional?

betsyknoke: I had to comment that out to get my version to run. I've fixed it now.

AfanasievAnton (approved):

PR #263, opened 2024-02-28
MTSDK 312-Error_Event_Is_Dispatched_When_CA_Set_Before_Session_Connected

ksw5:
- Implement MAX_CUSTOM_DATA_BYTES_UNSET for clear indication of unset size limits.
- Add maxCustomDataBytes setter to validate and potentially clear attributes if size limit is exceeded upon configuration.
- Simplify isCustomAttributesValid method by extracting the size exceeded check to separate function maybeReportFailure.
- Create maybeReportFailure function for a centralized CA size check.
- Reset maxCustomDataBytes upon session closure.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/CustomAttributesStoreImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/CustomAttributesStoreImpl.kt
@@ -16,12 +16,12 @@ internal class CustomAttributesStoreImpl(
     private var customAttributes: MutableMap<String, String> = mutableMapOf()
     var maxCustomDataBytes: Int = MAX_CUSTOM_DATA_BYTES_UNSET
         internal set(value) {
-            if (!value.isUnset()) {
+            field = value
+            if (maxCustomDataBytes.isSet()) {
                 if (maybeReportFailure(customAttributes)) {
                     customAttributes.clear()
                 }
             }
-            field = value
         }
 
     internal var state: State = State.PENDING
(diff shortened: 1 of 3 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/CustomAttributesStoreImpl.kt
```diff
@@ -30,10 +40,14 @@ internal class CustomAttributesStoreImpl(
     }
 
     private fun isCustomAttributesValid(customAttributes: Map<String, String>): Boolean {
-        if (customAttributes.isEmpty() || this.customAttributes == customAttributes) {
-            log.w { LogMessages.CUSTOM_ATTRIBUTES_EMPTY_OR_SAME }
-            return false
-        } else if (isSizeExceeded(customAttributes)) {
+        return maxCustomDataBytes.isUnset() || (
```
AfanasievAnton, line 43: The return condition of this function looks quite difficult for reading. 
I think we can sacrifice a couple of lines of code to make it more readable for human eye

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/CustomAttributesStoreImpl.kt
```diff
@@ -30,10 +40,14 @@ internal class CustomAttributesStoreImpl(
...
+            !(customAttributes.isEmpty() || this.customAttributes == customAttributes) &&
+                !maybeReportFailure(customAttributes)
+            )
+    }
+
+    private fun maybeReportFailure(customAttributes: Map<String, String>): Boolean {
+        return isSizeExceeded(customAttributes).also {
```
AfanasievAnton, line 50: code inside `also` scope will be executed each and every time leading to report Event.Error on every invocation of  `maybeReportFailure`.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/CustomAttributesStoreImpl.kt
```diff
@@ -7,12 +7,22 @@ import com.genesys.cloud.messenger.transport.util.logs.LogMessages
...
+
 internal class CustomAttributesStoreImpl(
     private val log: Log,
     private val eventHandler: EventHandler,
 ) : CustomAttributesStore {
     private var customAttributes: MutableMap<String, String> = mutableMapOf()
-    var maxCustomDataBytes: Int = 0
+    var maxCustomDataBytes: Int = MAX_CUSTOM_DATA_BYTES_UNSET
+        internal set(value) {
+            if (!value.isUnset()) {
```
AfanasievAnton, line 19: Something does not properly work in this `if` condition. 

1.  `maybeReportFailure` relies on the value of maxCustomDataBytes when size is validated. But the maxCustomDataBytes field is set only in line 24 (field = value). This leads to situation where isSizeValid return false while it is not correct.

AfanasievAnton (changes requested):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/CustomAttributesStoreImpl.kt
```diff
@@ -30,10 +40,27 @@ internal class CustomAttributesStoreImpl(
     }
 
     private fun isCustomAttributesValid(customAttributes: Map<String, String>): Boolean {
+        // Check if the size limit is unset, which means any size is valid
```
AfanasievAnton, line 43: I would say that comments here are redundant. Usually, we want to leave comments  when we have a weird logic or some workaround.

AfanasievAnton (changes requested):

AfanasievAnton (approved):

PR #295, opened 2024-07-11
Feature/mtsdk 340 create automation tests on i os

ksw5:
-Create iOS automation tests


iosApp/iosAppTests/iosAppTests.swift
```diff
@@ -596,12 +596,98 @@ class iosAppTests: XCTestCase {
...
         testController.verifyReceivedMessage(expectedMessage: receivedMessageText)
-
+        
         // Disconnect the conversation for the agent and disconnect the session.
         ApiHelper.shared.sendConnectOrDisconnect(conversationInfo: conversationInfo, connecting: false, wrapup: true)
         testController.disconnectMessenger()
     }
-
+    
+    func testStepUpAuthentication() {
```
mmorehou, line 605: A possible issue with running this test with the full suite is that the suite runs tests in alphabetical order. This test would be ran very late and at that point, the auth token this test requires may expire before the test runs. Another test adds the prefix `AAA_` to the test name to mitigate this.

mmorehou:
I see node_modules showing up in this PR. However I'm not sure what node is used for in this feature.

iosApp/iosAppTests/iosAppTests.swift
ksw5: Ok, I will make this change. Thank you.

mmorehou (approved):

PR #307, opened 2025-02-03
MTSDK-524 Public api to sync device token

AfanasievAnton:
- Add utility function in MessengerTransportSDK that  creates and return PushService.
- Add PushService.kt interface as public api.
- Add synchronize() function to PushService
- Add PushServiceImpl.kt to handle push related functionality.
- Add NotificationProvider.kt to represent possible push notification providers on mobile.
- Add KDoc


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/push/PushService.kt
```diff
@@ -0,0 +1,18 @@
...
+
+    /**
+     * Synchronizes the device token with the specified notification provider.
+     *
+     * @param deviceToken The device token for push notifications.
+     * @param notificationProvider The type of notification provider to use.
+     *
+     * @throws Exception if synchronization fails.
+     */
+    @Throws(Exception::class)
```
janoskernacsgenesys, line 16: This Exception class seems to be too generic. Will we specify a more specific exception class later?

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/push/NotificationProvider.kt
```diff
@@ -0,0 +1,12 @@
+package com.genesys.cloud.messenger.transport.push
+
+/**
+ * Represents push notification providers for mobile platforms.
+ *
+ * @property APNS Apple Push Notification Service.
+ * @property FCM Firebase Cloud Messaging.
+ */
+enum class NotificationProvider {
```
KalmanPerkelGenesys, line 9: Consider renaming `NotificationProvider` to `PushProvider` since not all push messages result in user-visible notifications

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/push/PushService.kt
AfanasievAnton: Good call. 
It is still not 100% clear what are going to be the types of errors Thandora (our backend push service) will return. Therefore, we use generic Exception now, with chance that it will be adjusted in the future.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/push/NotificationProvider.kt
AfanasievAnton: Done

janoskernacsgenesys (approved):

KalmanPerkelGenesys (approved):

PR #316, opened 2025-02-13
techdebt/MTSDK-541_no_feedback_when_detach_is_called_with_an_invalid_id

ksw5:
- Logs an error when detach is called with an invalid attachment ID.

- Update `when detach() on not processed attachment` test.

### DIFF
--- a/transport/src/androidUnitTest/kotlin/transport/core/AttachmentHandlerImplTest.kt
+++ b/transport/src/androidUnitTest/kotlin/transport/core/AttachmentHandlerImplTest.kt
@@ -293,12 +293,14 @@ internal class AttachmentHandlerImplTest {
 
     @Test
     fun `when detach() on not processed attachment`() {
-        val result = subject.detach(TestValues.Token, "not processed attachment id")
+        val invalidAttachmentId = "not processed attachment id"
 
-        verify {
-            listOf(mockAttachmentListener) wasNot Called
+        val exception = assertFailsWith<IllegalArgumentException> {
+            subject.detach(TestValues.Token, invalidAttachmentId)
         }
-        assertThat(result).isNull()
+        assertThat(exception.message).isEqualTo("Detach failed: Invalid attachment ID ($invalidAttachmentId)")
+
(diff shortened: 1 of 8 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/AttachmentHandlerImpl.kt
```diff
@@ -85,6 +85,11 @@ internal class AttachmentHandlerImpl(
     }
 
     override fun detach(token: String, attachmentId: String): DeleteAttachmentRequest? {
```
AfanasievAnton, line 87: - Add @Throws annotation to detach() function. 
- Also update the @Throws annotation of the MessagingClient.detach() as right now it's signature say that it throws ONLY IllegalStateException
- The comment is redundant.
- Create and use ErrorMessage.detachFailed() as the message parameter when you throw IllegalArgumentException instead of String. Reuse it later on in your unit test.

transport/src/androidUnitTest/kotlin/transport/core/AttachmentHandlerImplTest.kt
```diff
@@ -293,12 +293,14 @@ internal class AttachmentHandlerImplTest {
 
     @Test
     fun `when detach() on not processed attachment`() {
-        val result = subject.detach(TestValues.Token, "not processed attachment id")
+        val invalidAttachmentId = "not processed attachment id"
```
AfanasievAnton, line 296: - Rename invalidAttachmentId to givenAttachmentId to indicate that this is an input value we use for testing.
- Use TestValues.DEFAULT_STRING instead of hardcoded  "not processed attachment id".
- I would like to see a unit test in MCAttachmentTests ->  `when detach() non existing attachmentId` updated to reflect the change. Initial assumption of this test was that when detach non existing attachment attachmentHandler will return null. But now, it throws an exception.

AfanasievAnton (changes requested):

AfanasievAnton (approved):
Nice and clean!

PR #319, opened 2025-02-18
Feature/mtsdk 529 implement register device token

AfanasievAnton:
- Implement registerDeviceToken api call
- Move urls from Configuration.kt into separate Urls.kt file.
- Update dependencies to use Urls.kt instead of Configuration.kt
- Add log message to print the actual diff when comparing 2 PushConfig objects
- Add RegisterDeviceTokenRequestBody.kt to simplify serialization of deviceToken json body needed for registerDeviceToken request
- Update the DEVICE_TOKEN_EXPIRATION_IN_SECONDS to 30 days
- Add/Update unit tests

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/Urls.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/Urls.kt
@@ -5,6 +5,9 @@ import io.ktor.http.URLBuilder
 import io.ktor.http.Url
 import io.ktor.http.path
 
+private const val BASE_WEBDEPLOYMENTS_PATH = "api/v2/webdeployments"
+private const val BASE_WEBMESSAGING_PATH = "api/v2/webmessaging"
+
 internal class Urls(val domain: String, val deploymentId: String) {
 
     internal val webSocketUrl: Url by lazy {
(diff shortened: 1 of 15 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/Urls.kt
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    internal val refreshAuthTokenUrl: Url by lazy {
+        URLBuilder("https://api.$domain").apply {
+            path("api/v2/webdeployments/token/refresh")
+        }.build()
+    }
+
+    internal val registerDeviceToken: (String, String) -> Url = { deploymentId, token ->
+        URLBuilder("https://api.$domain").apply {
```
janoskernacsgenesys, line 49: This repeating string could be extracted to a const val like BASE_API_URL

ksw5 (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/Urls.kt
janoskernacsgenesys: same for the others (api-cdn, websocket)

janoskernacsgenesys (approved):

PR #343, opened 2025-04-11
techdebt/MTSDK-555_Flaky_whenUploadFile_fails_with_ResponseError

ksw5:
- Replaced newSingleThreadContext with Dispatchers.Unconfined for stability

### DIFF
--- a/transport/src/androidUnitTest/kotlin/transport/core/AttachmentHandlerImplTest.kt
+++ b/transport/src/androidUnitTest/kotlin/transport/core/AttachmentHandlerImplTest.kt
@@ -34,9 +34,9 @@ import io.mockk.mockk
 import io.mockk.slot
 import io.mockk.spyk
 import io.mockk.verify
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.newSingleThreadContext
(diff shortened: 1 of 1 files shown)


AfanasievAnton (approved):
That should do the trick

transport/src/androidUnitTest/kotlin/transport/core/AttachmentHandlerImplTest.kt
```diff
@@ -62,8 +62,7 @@ internal class AttachmentHandlerImplTest {
     private val givenUploadSuccessEvent = uploadSuccessEvent()
 
     @ExperimentalCoroutinesApi
-    private val threadSurrogate = newSingleThreadContext("main thread")
-
+    private val dispatcher: CoroutineDispatcher = Dispatchers.Unconfined
```
janoskernacsgenesys, line 65: I suggest UnconfinedTestDispatcher instead. Dispatchers.Unconfined. 

Dispatchers.Unconfined starts executing a coroutine in the current thread (the thread that launched the coroutine). However, it does not confine the coroutine to that thread. When the coroutine suspends, it can resume on any thread. Tests need to be deterministic and repeatable, and Dispatchers.Unconfined introduces too much uncertainty

UnconfinedTestDispatcher is designed specifically for testing. It executes coroutines immediately in the current thread until the first suspension point. After suspension, it will resume on the same thread.

AfanasievAnton (approved):

janoskernacsgenesys (approved):

PR #346, opened 2025-04-22
feature/MTSDK-660-normalize-c-c

ksw5:
- Added new Message.Type values for Card and Carousel
- Updated toMessage() to combine cards and carousel into a single list
- Simplified toMessageType() to check for quick replies or carousel
- Parsed CardContent and CarouselContent into Message.Card objects
- Adjusted serializers and models to support new message types

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/Message.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/Message.kt
@@ -51,6 +51,8 @@ data class Message(
         Text,
         Event,
         QuickReply,
+        Card,
+        Carousel,
         Unknown,
     }
 
@@ -106,6 +108,9 @@ data class Message(
         val contentType: Type,
         val attachment: Attachment? = null,
         val buttonResponse: ButtonResponse? = null,
+        val card: Card? = null,
+        val carousel: Carousel? = null,
+
     ) {
         @Serializable
         enum class Type {
@@ -141,4 +146,24 @@ data class Message(
             Unknown,
         }
     }
+    @Serializable
+    data class Card(
+        val title: String,
+        val description: String,
+        val imageUrl: String,
+        val actions: List<Action>
+    ) {
+        @Serializable
+        data class Action(
+            val type: String,
+            val title: String? = null,
+            val url: String? = null,
+            val payload: String? = null,
+        )
+    }
+
+    @Serializable
+    data class Carousel(
+        val cards: List<Card>
+    )
 }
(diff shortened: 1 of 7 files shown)


janoskernacsgenesys (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/extensions/MessageExtension.kt
```diff
@@ -24,15 +24,19 @@ internal fun List<StructuredMessage>.toMessageList(): List<Message> =
 
 internal fun StructuredMessage.toMessage(): Message {
     val quickReplies = content.toQuickReplies()
+    val cards = content.toCards()
+    val carousel = content.toCarousel()
+    val cardsAndCarousel = cards + carousel
```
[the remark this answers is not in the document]
ksw5: It seems that it isn't possible.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/Message.kt
```diff
@@ -51,6 +51,8 @@ data class Message(
         Text,
         Event,
         QuickReply,
+        Card,
```
AfanasievAnton, line 54: Seems like `Card` type is never used. Are we going to leave only Carousel option? If yes, we should remove it. Another option is to have Message.Type.Cards (combination of carousel (multiple cards) and single card).

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/Message.kt
```diff
@@ -106,6 +108,9 @@ data class Message(
         val contentType: Type,
         val attachment: Attachment? = null,
         val buttonResponse: ButtonResponse? = null,
+        val card: Card? = null,
```
AfanasievAnton, line 111: Do we really need card and carousel here or we can reuse buttonResponse?

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/Message.kt
```diff
@@ -141,4 +146,24 @@ data class Message(
...
+    data class Card(
+        val title: String,
+        val description: String,
+        val imageUrl: String,
+        val actions: List<Action>
+    ) {
+        @Serializable
+        data class Action(
+            val type: String,
+            val title: String? = null,
```
AfanasievAnton, line 159: I think there is no such thing as Action.Title but instead it is Action.text

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/Message.kt
ksw5: Thanks! I checked the inbound response for both Card and Carousel clicks, and it looks like they always come in as ButtonResponse. Just to confirm — should we keep card and carousel only for outbound messages, and rely on buttonResponse for all inbound button clicks?

PR #351, opened 2025-05-13
MTSDK - 545 Implement EncryptedVault

yarin-sason:


transport/src/androidUnitTest/kotlin/transport/util/EncryptedVaultTest.kt
```diff
@@ -0,0 +1,134 @@
...
+    @Test
+    fun `test store delegates to InternalVault`() {
+        // Given an EncryptedVault instance
+        val encryptedVault = EncryptedVault(testKeys)
+        
+        // When storing a key-value pair
+        encryptedVault.store(testKey, testValue)
+        
+        // Then it should delegate to InternalVault
+        verify { anyConstructed<InternalVault>().store(testKey, testValue) }
```
AfanasievAnton, line 81: This one also something new to me. I understand the idea behind it. Just wondering what is the benefit of it. 
anyConstructed<InternalVault>(): Refers to any instance of the InternalVault class that was constructed during the test, not just a specific instance. I guess in the context of this particular use case we are not expecting to have multiple instances of InternalVaults constructed so it is fine. But we should use it carefully.

transport/src/androidUnitTest/kotlin/transport/core/MessengerTransportSDKTest.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    @Test
+    fun `should use provided vault regardless of encryptedVault setting`() {
+        // Given
+        val configuration = Configuration(
+            deploymentId = "testDeploymentId",
+            domain = "testDomain",
+            encryptedVault = true // This would normally create an EncryptedVault
+        )
+        val customVault = DefaultVault() // But we're providing a DefaultVault
```
AfanasievAnton, line 79: You can use FakeVault to emphasize that this is actually a custom implementation

transport/src/commonTest/kotlin/com/genesys/cloud/messenger/transport/utility/MockVault.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.genesys.cloud.messenger.transport.util.TOKEN_KEY
+import com.genesys.cloud.messenger.transport.util.AUTH_REFRESH_TOKEN_KEY
+import com.genesys.cloud.messenger.transport.util.WAS_AUTHENTICATED
+import com.genesys.cloud.messenger.transport.util.Vault
+
+/**
+ * A mock implementation of the Vault interface for testing purposes.
+ * This implementation stores values in a simple in-memory map.
+ */
+class MockVault(
```
AfanasievAnton, line 13: Check FakeVault. It looks pretty similar to that :)

AfanasievAnton (approved):

transport/src/androidUnitTest/kotlin/transport/util/EncryptedVaultTest.kt
yarin-sason: The reason its usefully here is because the internal Vault is a private member so it's not accessible otherwise ,on the other hand i now wondering if we should even have that test, because i guess that if we want to handle store in a different manner then using the internal vault it should be also fine (and not to fore here the use of the internal vault)

transport/src/androidUnitTest/kotlin/transport/util/EncryptedVaultTest.kt
```diff
@@ -0,0 +1,134 @@
...
+    private val mockSharedPreferencesEditor = mockk<SharedPreferences.Editor>(relaxed = true)
+
+    // Test data
+    private val testKeys = Vault.Keys(
+        vaultKey = TestValues.VAULT_KEY,
+        tokenKey = TestValues.TOKEN_KEY,
+        authRefreshTokenKey = TestValues.AUTH_REFRESH_TOKEN_KEY,
+        wasAuthenticated = TestValues.WAS_AUTHENTICATED
+    )
+    private val testKey = "test_key"
```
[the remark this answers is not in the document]
yarin-sason: Thanks for the comment. I've rephrased it. There are some variables that are used both as "given" values and "expected" values, so I left them without a prefix. I hope that makes sense.

transport/src/androidMain/kotlin/com/genesys/cloud/messenger/transport/util/EncryptedVault.android.kt
```diff
@@ -0,0 +1,47 @@
...
+actual class EncryptedVault actual constructor(keys: Keys) :
+    Vault(keys) {
+    private val sharedPreferences: SharedPreferences
+    private val internalVault: InternalVault
+
+    init {
+        if (context == null) {
+            throw IllegalStateException("Must set EncryptedVault.context before instantiating.")
+        }
+        sharedPreferences = context!!.getSharedPreferences(keys.vaultKey, Context.MODE_PRIVATE)
```
janoskernacsgenesys, line 22: Please avoid double exclamation marks. This is always a sign of a wrong nullability usage in Kotlin. 
I advise to use weak reference to the context. 

OR even to use something other storage type which does not need context

transport/src/iosMain/kotlin/com/genesys/cloud/messenger/transport/util/EncryptedVault.ios.kt
```diff
@@ -0,0 +1,38 @@
+package com.genesys.cloud.messenger.transport.util
+
+import com.genesys.cloud.messenger.transport.core.InternalVault
+
+/**
+ * iOS implementation of EncryptedVault that wraps the iOS default vault (Keychain).
+ */
+actual class EncryptedVault actual constructor(keys: Keys) :
```
janoskernacsgenesys, line 8: Would the following be simpler and still working? 

```suggestion
actual class EncryptedVault actual constructor(keys: Keys) : InternalVault(keys)
```

transport/src/androidMain/kotlin/com/genesys/cloud/messenger/transport/util/EncryptedVault.android.kt
janoskernacsgenesys: Well there is not really possible to set up any kind of device storage without a Context

transport/src/androidMain/kotlin/com/genesys/cloud/messenger/transport/util/EncryptedVault.android.kt
```diff
@@ -0,0 +1,47 @@
...
+import android.content.SharedPreferences
+import com.genesys.cloud.messenger.transport.core.InternalVault
+import java.lang.ref.WeakReference
+
+/**
+ * Android implementation of EncryptedVault that uses Android KeyStore
+ * for the encryption keys and stores the encrypted data in SharedPreferences.
+ *
+ */
+actual class EncryptedVault actual constructor(keys: Keys) :
```
janoskernacsgenesys, line 13: Well, this is just a wrap around the InrternalVault. 
Would you think it would be better just to move the content of InternalVault here?

transport/src/androidMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalVault.kt
```diff
@@ -0,0 +1,215 @@
...
+     */
+    private fun generateSecretKey(): SecretKey {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            val keyGenParameterSpec = KeyGenParameterSpec.Builder(
+                keyAlias,
+                KeyProperties.PURPOSE_ENCRYPT or KeyProperties.PURPOSE_DECRYPT
+            )
+                .setBlockModes(KeyProperties.BLOCK_MODE_GCM)
+                .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_NONE)
+                .setKeySize(256)
```
janoskernacsgenesys, line 200: I would suggest to add these as well: 

```
         .setRandomizedEncryptionRequired(true)
         .setUserAuthenticationRequired(false)
```

transport/src/iosMain/kotlin/com/genesys/cloud/messenger/transport/util/EncryptedVault.ios.kt
yarin-sason: umm not sure I understand , the Encrypted the actual class of the expect class 
image

so it cannot be inherited from InternalVault

transport/src/androidMain/kotlin/com/genesys/cloud/messenger/transport/util/EncryptedVault.android.kt
yarin-sason: something like that ? 

image
we already use weak reference for the context as it set to a companion object similar to the implementation of Default Vault

transport/src/androidMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalVault.kt
yarin-sason: Those are the default settings anyway do you think it will be better to add for clarity ?

transport/src/androidMain/kotlin/com/genesys/cloud/messenger/transport/util/EncryptedVault.android.kt
AfanasievAnton: > Please avoid double exclamation marks. This is always a sign of a wrong nullability usage in Kotlin. I advise to use weak reference to the context.
> 
> ~OR even to use something other storage type which does not need context~

We check the context for nullability. So in this case I am ok with using `!!`. 
We could do something like that:
`
context?.let {
    sharedPreferences = it.getSharedPreferences(keys.vaultKey, Context.MODE_PRIVATE)
} ?: throw IllegalStateException("Must set EncryptedVault.context before instantiating.")
`


 but then it mess around the sharedPrefs and internalVault fields as they are declared final. 
 
 So the only way around is to have the store the context in non nullable local field like that:
 `
    val ctx = context ?: throw IllegalStateException("Must set EncryptedVault.context before instantiating.")
    sharedPreferences = ctx.getSharedPreferences(keys.vaultKey, Context.MODE_PRIVATE)
    internalVault = InternalVault(keys.vaultKey, sharedPreferences)
`

transport/src/androidMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalVault.kt
janoskernacsgenesys: If these are the default values, then OK

janoskernacsgenesys (approved):

PR #353, opened 2025-05-14
Task/mtsdk 655 remove okta properties

ksw5:
Address security concerns by removing hardcoded sensitive information from the public transport-mobile-sdk repository.

Migrated okta.properties and Okta.plist secrets to Secret Stash
Updated Jenkinsfile to securely fetch secrets at runtime
Replaced plain text credentials with secure environment variables
Validated successful Android and iOS builds using new secure flow


AfanasievAnton (approved):

janoskernacsgenesys (approved):

Jenkinsfile
```diff
@@ -3,88 +3,126 @@ def isMainBranch() { env.BRANCH_NAME == 'main' }
...
+                        }
+                    }
+                    steps {
+                        script {
+                            def pipelineLibrary = library('pipeline-library').com.genesys.jenkins
+                            def testing = pipelineLibrary.Testing.new()
+                            oktaproperties = testing.getSecretStashSecret(
+                                'dev',
+                                'us-east-1',
+                                'mobiledx-ios',
```
mobichel, line 51: I saw comment in the chat:
`I think we should get the transportsdk stash here instead of mobiledx-ios.`
and we still have `mobiledx-ios`.
What was the resolution?

mobichel (changes requested):
Please clarify why we're using `transportsdk` stash name?

Jenkinsfile
AfanasievAnton: > I saw comment in the chat: `I think we should get the transportsdk stash here instead of mobiledx-ios.` and we still have `mobiledx-ios`. What was the resolution?

I agree with you. This is a temporary workaround.
We decided to handle it separately in the next sprint: https://inindca.atlassian.net/browse/MTSDK-688
For now we can fetch okta.properites from mobiledx-ios group.

abarshev (approved):
LGTM given https://inindca.atlassian.net/browse/MTSDK-688

mobichel (approved):

PR #355, opened 2025-05-20
Hide google-service.json behind secret stash

ksw5:
Hide google-service.json behind secret stash

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -48,7 +48,7 @@ pipeline {
                             oktaproperties = testing.getSecretStashSecret(
                                 'dev',
                                 'us-east-1',
-                                'mobiledx-ios',
+                                'transportsdk',
                                 'okta-properties',
                                 env.WORKSPACE
                             )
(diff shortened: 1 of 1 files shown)


janoskernacsgenesys (approved):

Jenkinsfile
```diff
@@ -33,34 +34,57 @@ pipeline {
...
-                    }
+            agent {
+                node {
+                    label 'dev_mesos_v2'
+                    customWorkspace "jenkins-mtsdk-${currentBuild.number}"
+                }
+            }
+            steps {
+                script {
+                    def pipelineLibrary = library('pipeline-library').com.genesys.jenkins
```
AfanasievAnton, line 46: Looks like we are obtaining the reference to pipelinLibrary twice. Once for oktaProperties, second time for google-services. 
Is it possible to somehow combine them into single step, like "retrieve values from secret stash" or something ?

AfanasievAnton (approved):

PR #356, opened 2025-05-21
MTSDK-679 Support Identity stitching error

janoskernacsgenesys:
Handle the new error + update the tests

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ErrorCode.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ErrorCode.kt
@@ -33,6 +33,7 @@ sealed class ErrorCode(val code: Int) {
     data object RefreshAuthTokenFailure : ErrorCode(6003)
     data object HistoryFetchFailure : ErrorCode(6004)
     data object ClearConversationFailure : ErrorCode(6005)
+
     // Push
     data object DeviceTokenOperationFailure : ErrorCode(6020)
     data object DeviceAlreadyRegistered : ErrorCode(6021)
@@ -41,6 +42,7 @@ sealed class ErrorCode(val code: Int) {
     data object DeviceRegistrationFailure : ErrorCode(6024)
     data object DeviceUpdateFailure : ErrorCode(6025)
     data object DeviceDeleteFailure : ErrorCode(6026)
+    data object IdentityResolutionDisabled : ErrorCode(6027)
 
     data class RedirectResponseError(val value: Int) : ErrorCode(value)
     data class ClientResponseError(val value: Int) : ErrorCode(value)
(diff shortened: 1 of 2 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ErrorCode.kt
```diff
@@ -139,7 +145,8 @@ internal fun ErrorCode.toCorrectiveAction(): CorrectiveAction = when (this.code)
     ErrorCode.AuthFailed.code,
     ErrorCode.AuthLogoutFailed.code,
     ErrorCode.RefreshAuthTokenFailure.code,
-    -> CorrectiveAction.ReAuthenticate
+        -> CorrectiveAction.ReAuthenticate
+
```
yarin-sason, line 149: syntext ?

yarin-sason (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ErrorCode.kt
```diff
@@ -41,6 +42,7 @@ sealed class ErrorCode(val code: Int) {
     data object DeviceRegistrationFailure : ErrorCode(6024)
     data object DeviceUpdateFailure : ErrorCode(6025)
     data object DeviceDeleteFailure : ErrorCode(6026)
+    data object IdentityResolutionDisabled : ErrorCode(6027)
```
AfanasievAnton, line 45: This one is never used (except of test) So we should either update the PushErrorResponse.toErrorCode() to return this error code when we are doing mapping. Or completely delete it and treat any "identity.resolution.disabled" as ErrorCode.DeviceRegistrationFailure. 
I think  option 2 is better because from Thandora perspective they define that this error can happen **only** during device registration.

AfanasievAnton (commented):
Can we please add one more assertion in ErrorCodeTest.`when PushErrorResponse toErrorCode`()
something like: assertThat(pushErrorResponseWith("identity.resolution.disabled").toErrorCode()).isEqualTo(ErrorCode. DeviceRegistrationFailure)

AfanasievAnton (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ErrorCode.kt
janoskernacsgenesys: This is just a result of autoformat at commit. Not related to the main task

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ErrorCode.kt
janoskernacsgenesys: Fixed

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ErrorCode.kt
janoskernacsgenesys: Is this an issue @yarin-sason?

AfanasievAnton (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ErrorCode.kt
yarin-sason: I think that's fine if it part of the autoformating

AfanasievAnton (approved):

PR #361, opened 2025-06-03
MTSDK-743 Add Client Type and Version Parameters to Configuration Class

yarin-sason:
https://inindca.atlassian.net/browse/MTSDK-743
(branch name was set to GMMS instead MTSDK by accident) 

- Added ConfigurationFactory object with a create() function that handles configuration creation including the messenger application parameter.  
- Introduced ApplicationType enum to distinguish between TRANSPORT and MESSENGER_SDK types
- Added validation to ensure messengerSDKVersion is provided when using MESSENGER_SDK type
- Implemented proper application string formatting:
  - For Messenger SDK: "MessengerSDK-{version}/TransportSDK-{version}"
  - For Transport SDK: "TransportSDK-{version}"
- Default application type is set to TRANSPORT for backward compatibility


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ConfigurationFactory.kt
```diff
@@ -0,0 +1,44 @@
+package com.genesys.cloud.messenger.transport.core
+
+object ConfigurationFactory {
```
AfanasievAnton, line 3: Lets name it InternalConfigurationFactory and leave a KDoc highlighting that this object is intended for internal usage only.

transport/src/commonTest/kotlin/com/genesys/cloud/messenger/transport/ConfigurationFactoryTest.kt
```diff
@@ -0,0 +1,65 @@
+package com.genesys.cloud.messenger.transport
+
+import com.genesys.cloud.messenger.transport.core.ApplicationType
+import com.genesys.cloud.messenger.transport.core.ConfigurationFactory
+import com.genesys.cloud.messenger.transport.core.MessengerTransportSDK
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+class ConfigurationFactoryTest {
```
AfanasievAnton, line 10: We are missing the testing of the default value of `application` parameter. You do test that value is set to Transport, but only when the parameter is explicitly set. So I am missing somethig like that:
```
    fun `when creating configuration with TRANSPORT type it should format application parameter correctly`() {
        val config = ConfigurationFactory.create(
            deploymentId = "test-deployment",
            domain = "test.com"
        )
                assertEquals(
            "TransportSDK-${MessengerTransportSDK.sdkVersion}",
            config.application
        )
```

yarin-sason: have a look at line 54 , I think that you ment for this test case right?

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ConfigurationFactory.kt
yarin-sason: done

AfanasievAnton (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ApplicationType.kt
```diff
@@ -0,0 +1,6 @@
+package com.genesys.cloud.messenger.transport.core
+
+enum class ApplicationType {
+    TRANSPORT,
```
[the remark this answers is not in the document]
yarin-sason: Make sense, I will update

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalConfigurationFactory.kt
```diff
@@ -0,0 +1,48 @@
...
+     * @param reconnectionTimeoutInSeconds period of time during which Transport will try to reconnect.
+     * @param autoRefreshTokenWhenExpired indicates if Transport should auto refresh auth token if expired.
+     * @param encryptedVault indicates if encrypted vault should be used.
+     * @return Configuration instance with proper application parameter formatting.
+     */
+    fun create(
+        deploymentId: String,
+        domain: String,
+        applicationType: ApplicationType = ApplicationType.TRANSPORT_SDK,
+        messengerSDKVersion: String? = null,
```
mobichel, line 23: I would suggest keep the name of this field more generic and not bound to the messenger. 
For example `applicationVersion` in case if we add more application types it still will be actual.

mobichel: I would suggest to make this field mandatory, since it will be used only by Messenger, we should force it to set the version.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalConfigurationFactory.kt
```diff
@@ -0,0 +1,48 @@
...
+            domain = domain,
+            logging = logging,
+            reconnectionTimeoutInSeconds = reconnectionTimeoutInSeconds,
+            autoRefreshTokenWhenExpired = autoRefreshTokenWhenExpired,
+            encryptedVault = encryptedVault
+        )
+
+        config.application = when (applicationType) {
+            ApplicationType.TRANSPORT_SDK -> "TransportSDK-${MessengerTransportSDK.sdkVersion}"
+            ApplicationType.MESSENGER_SDK -> "MessengerSDK-$messengerSDKVersion/TransportSDK-${MessengerTransportSDK.sdkVersion}"
```
mobichel, line 44: I think we can make the second case more generic:
```
else -> "${applicationType}-$applicationVersion/TransportSDK-${MessengerTransportSDK.sdkVersion}"
```
Then we don't have to maintain this each time we're adding or changing `ApplicationType` enum.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalConfigurationFactory.kt
```diff
@@ -0,0 +1,48 @@
...
+        deploymentId: String,
+        domain: String,
+        applicationType: ApplicationType = ApplicationType.TRANSPORT_SDK,
+        messengerSDKVersion: String? = null,
+        logging: Boolean = false,
+        reconnectionTimeoutInSeconds: Long = 60 * 5,
+        autoRefreshTokenWhenExpired: Boolean = true,
+        encryptedVault: Boolean = false
+    ): Configuration {
+        require(!(applicationType == ApplicationType.MESSENGER_SDK && messengerSDKVersion == null)) {
```
mobichel, line 29: See my comment above, I thinks we should force it on level of function argument.

mobichel (changes requested):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalConfigurationFactory.kt
yarin-sason: ok I will change that accordingly, for both fields?

mobichel: As agreed, applicationType will also not have the default value.

androidComposePrototype/src/main/java/com/genesys/cloud/messenger/androidcomposeprototype/ui/testbed/TestBedViewModel.kt
```diff
@@ -83,12 +84,20 @@ class TestBedViewModel : ViewModel(), CoroutineScope {
...
+        val mmsdkConfiguration = InternalConfigurationFactory.create(
             deploymentId = deploymentId.ifEmpty { BuildConfig.DEPLOYMENT_ID },
             domain = region.ifEmpty { BuildConfig.DEPLOYMENT_DOMAIN },
             logging = true,
-            encryptedVault = true
+            encryptedVault = true,
+            applicationType = ApplicationType.MESSENGER_SDK,
+            applicationVersion = "3.3.3"
         )
+//        val mmsdkConfiguration = Configuration(
```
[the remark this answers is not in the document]
yarin-sason: pushed it by mistake one sec

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ApplicationType.kt
```diff
@@ -0,0 +1,10 @@
+package com.genesys.cloud.messenger.transport.core
+
+enum class ApplicationType {
+    TRANSPORT_SDK {
+        override fun toString(): String = "TransportSDK"
```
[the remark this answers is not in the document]
yarin-sason: look bit more cleaner to me for use , but I can use attribute instead

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalConfigurationFactory.kt
```diff
@@ -9,7 +9,7 @@ object InternalConfigurationFactory {
      * @param deploymentId the ID of the Genesys Cloud Messenger deployment.
      * @param domain the regional base domain address for a Genesys Cloud Web Messaging service.
      * @param applicationType the type of SDK being used (Transport or Messenger). Defaults to TRANSPORT.
```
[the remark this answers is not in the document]
yarin-sason: np

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ApplicationType.kt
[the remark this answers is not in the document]
mobichel: Can you do this instead:

```
enum class ApplicationType(val title: String) {
    TRANSPORT_SDK("TransportSDK"),
    MESSENGER_SDK("MessengerSDK");
    
    override fun toString(): String = title
}
```

?

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalConfigurationFactory.kt
```diff
@@ -0,0 +1,45 @@
...
+     * @param reconnectionTimeoutInSeconds period of time during which Transport will try to reconnect.
+     * @param autoRefreshTokenWhenExpired indicates if Transport should auto refresh auth token if expired.
+     * @param encryptedVault indicates if encrypted vault should be used.
+     * @return Configuration instance with proper application parameter formatting.
+     */
+    fun create(
+        deploymentId: String,
+        domain: String,
+        applicationType: ApplicationType,
+        applicationVersion: String?,
```
mobichel, line 23: `String?` is it still optional on purpose?

yarin-sason: Why you think it optional ? 
we leave the option to set it to applicationVersion = null

mobichel (approved):

AfanasievAnton (approved):

PR #385, opened 2025-07-27
MTSDK-759 Encryption Error When Switching Between Default and Encrypted Vault

yarin-sason:
That issue was appears when we try to decrypt regular string (the decrypt function expexct base64 but got regular string), and also the other direction not raised errors but return, weird responses.

the propose solution use different shred pref files for storing encrypted and regular key- value pairs for avoiding any potentially collisions, the two files will only papers in devices that switch between the two cases, for the most part vault will create one file 

image

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/EncryptedVault.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/EncryptedVault.kt
@@ -1,11 +1,13 @@
 package com.genesys.cloud.messenger.transport.util
 
+internal const val ENCRYPTED_VAULT_KEY = "com.genesys.cloud.messenger.encrypted"
+
 /**
  * A encrypted implementation of the [Vault].
  */
 expect class EncryptedVault(
     keys: Keys = Keys(
-        vaultKey = VAULT_KEY,
+        vaultKey = ENCRYPTED_VAULT_KEY,
         tokenKey = TOKEN_KEY,
         authRefreshTokenKey = AUTH_REFRESH_TOKEN_KEY,
         wasAuthenticated = WAS_AUTHENTICATED,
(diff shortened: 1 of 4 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/EncryptedVault.kt
```diff
@@ -1,11 +1,13 @@
...
 
+internal const val ENCRYPTED_VAULT_KEY = "com.genesys.cloud.messenger.encrypted"
+
 /**
  * A encrypted implementation of the [Vault].
  */
 expect class EncryptedVault(
     keys: Keys = Keys(
-        vaultKey = VAULT_KEY,
+        vaultKey = ENCRYPTED_VAULT_KEY,
```
AfanasievAnton, line 10: Happy to be wrong but, with this implementation switch between vault types will result in having different session tokens on each vault. Meaning we would not be able to resume an existing session stored with a different type of vault. So migration for existing clients might be an issue here. Is there a way to solve this bug without having 2 separate shared preferences files? Or at least synchronize between them?

AfanasievAnton: Common - no, but possible - definitely yes. Current customers that has DefaultVault and want to switch to EncryptedVault will face this problem. 
Customers that use default configuration == anyone who use MobileMessenger will have DefaultVault. I think at this point MobileMessenger does not expose option to configure this field as well as doesn't set it to useEncrypted = true. 
Customer can set custom keys, right. We can instruct customer that do migration from DefaultVault to EncryptedVault (or the other way around) use custom keys and decode/encode this keys values when necessary. 
Something like:
```
// Go to online decode/encode tool and use it to translate the old key values accordingly (maybe provide a tool to generate/convert this values as a utility function?) and then just update the keys.
messengerTransport = MessengerTransportSDK(
            mmsdkConfiguration, EncryptedVault(
                Vault.Keys(
                    vaultKey = "decoded vault key ",
                    tokenKey = "decoded toke key",
                    authRefreshTokenKey = "etc",
                    wasAuthenticated = "eyc",
                    pushConfigKey = "etc",
                )
            )
        )
```
Will it work?

yarin-sason: yes it also an option I've added out of the box solution that just automatically migrate a default vault to an encrypted vault if it detect one, as we discussed last week

janoskernacsgenesys (approved):

AfanasievAnton (approved):

PR #386, opened 2025-07-31
feature/MTSDK-574_handle-successful-response-from-shyrka

ksw5:
- Updated handleStructuredMessage() to handle card replies
- Added logic to update MessageStore with Message.Type.Cards
- Wrote a unit test to verify card parsing and MessageStore update flow

### DIFF
--- a/transport/src/androidUnitTest/kotlin/transport/core/MessageStoreTest.kt
+++ b/transport/src/androidUnitTest/kotlin/transport/core/MessageStoreTest.kt
@@ -603,20 +603,19 @@ internal class MessageStoreTest {
 
     @Test
     fun `when preparePostbackMessage is called then returns StructuredMessage with correct metadata`() {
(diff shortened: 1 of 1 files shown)


transport/src/androidUnitTest/kotlin/transport/util/Response.kt
```diff
@@ -77,6 +77,8 @@ internal object Response {
         """{"type":"response","class":"JwtResponse","code":200,"body":{"jwt":"some_jwt","exp":333}}"""
     fun clearConversationForbidden(errorMessage: String = "Presence events Conversation Clear are not supported") =
         """{"type":"response","class":"string","code":403,"body":"$errorMessage"}"""
+    const val onMessageWithPostbackCardReply =
```
yarin-sason, line 80: that's will be replaced by the actual response in a later story

janoskernacsgenesys (approved):

PR #399, opened 2025-08-26
MTSDK-858 Off-Schedule Dependency Updates - Major

janoskernacsgenesys:
- updated dependency versions
-necessary code adjustments
- added ktlint config


.editorconfig
```diff
@@ -0,0 +1,25 @@
+root = true
```
AfanasievAnton, line 1: @janoskernacsgenesys Isn't this set of rules disables practically most of our existing linter rules?

janoskernacsgenesys: I had a review, these are the minimal rules. We can decide which one to fix or keep.

PR #400, opened 2025-08-26
MTSDK-369 add the changes needed to the android TestBedViewModel.kt 

yarin-sason:
Add the changes needed to the android TestBedViewModel.kt  and remove Action for the card structure

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/extensions/MessageExtension.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/extensions/MessageExtension.kt
@@ -105,7 +105,7 @@ private fun List<StructuredMessage.Content>.toQuickReplies(): List<ButtonRespons
     }
 }
 
-private fun StructuredMessage.Content.Action.toMessageCardAction(): ButtonResponse =
+private fun StructuredMessage.Content.Action.toMessageCardAction(): ButtonResponse? =
     when {
         type.equals("Link", ignoreCase = true) -> ButtonResponse(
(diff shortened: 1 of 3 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/extensions/MessageExtension.kt
```diff
@@ -105,32 +105,29 @@ private fun List<StructuredMessage.Content>.toQuickReplies(): List<ButtonRespons
...
-private fun StructuredMessage.Content.Action.toMessageCardAction(): Message.Card.Action =
+private fun StructuredMessage.Content.Action.toMessageCardAction(): ButtonResponse =
     when {
-        type.equals("Link", ignoreCase = true) -> Message.Card.Action(
+        type.equals("Link", ignoreCase = true) -> ButtonResponse(
             type = "Link",
-            text = text,
-            url = url,
-            payload = null
+            text = text!!,
```
yarin-sason, line 112: @AfanasievAnton I decide to set in those case !!, if you think it risky we can set an empty string instead

AfanasievAnton: hmm...I think the StructuredMessage.Card.Action.Text field should not be nullable. So we should fix it on the parsing level.

Screenshot 2025-08-26 at 3 58 08 PM

AfanasievAnton (approved):

janoskernacsgenesys (approved):

PR #407, opened 2025-09-05
Improve unit test coverage

AfanasievAnton:
- Align naming conventions of UserTypingProviderTest and StateMachineTest with rest of the project
- Add coverage of UserTypingProvider default constructor.
- Add test that verifies that all branches of eligibleToRefresh() are tested.

### DIFF
--- a/transport/src/androidUnitTest/kotlin/transport/core/StateMachineTest.kt
+++ b/transport/src/androidUnitTest/kotlin/transport/core/StateMachineTest.kt
@@ -44,13 +44,13 @@ class StateMachineTest {
     }
 
     @Test
-    fun whenSubjectWasInitialized() {
+    fun `when subject was initialized`() {
         assertThat(subject.currentState).isIdle()
         assertTrue { subject.isInactive() }
     }
 
     @Test
-    fun whenOnConnectionOpened() {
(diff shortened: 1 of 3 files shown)


transport/src/androidUnitTest/kotlin/transport/core/StateMachineTest.kt
```diff
@@ -287,7 +287,7 @@ class StateMachineTest {
     }
 
     @Test
```
ksw5, line 289: Just wondering why this test doesn't follow the same naming structure of "when..."

transport/src/androidUnitTest/kotlin/transport/core/StateMachineTest.kt
```diff
@@ -287,7 +287,7 @@ class StateMachineTest {
...
-    fun verifyFullStateTransitionCycle() {
+    fun `verify full state transition cycle`() {
```
ksw5, line 290: Just wondering why this test doesn't follow the same naming structure of "when...". Is it on purpose?

ksw5 (approved):

KalmanPerkelGenesys (approved):

PR #411, opened 2025-09-10
bug/MTSDK-877_Sending-button-response-type-postback-result-in-error

ksw5:
Fix ButtonResponse type handling and update tests

- Normalize Postback → Button so messages don’t fail with 400
- Update toMessage mapping to correctly handle QuickReply vs Button
- Add tests for QuickReplyContent and ButtonResponseContent mapping

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/extensions/MessageExtension.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/extensions/MessageExtension.kt
@@ -31,7 +31,7 @@ internal fun StructuredMessage.toMessage(): Message {
         id = metadata["customMessageId"] ?: id,
         direction = if (isInbound()) Direction.Inbound else Direction.Outbound,
         state = Message.State.Sent,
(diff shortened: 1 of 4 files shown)


janoskernacsgenesys (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/extensions/MessageExtension.kt
```diff
@@ -92,17 +93,22 @@ private fun List<AttachmentContent>.toAttachments(): Map<String, Attachment> {
...
+    if (filteredQuickReply.isNotEmpty()) {
+        return filteredQuickReply.map {
+            it.quickReply.run { ButtonResponse(text, payload, "QuickReply") }
         }
+    }
 
-        else -> emptyList()
+    if (filteredButtonResponse.isNotEmpty()) {
+        return filteredButtonResponse
+            .mapNotNull { br ->
```
AfanasievAnton, line 105: please, do not use abbreviations in the project unless they are widely recognized. In this case "br" is quite confusing. Use "it" instead, or give it a more meaningful name.

AfanasievAnton (approved):
Approved, just make sure to fix small comment.

PR #479, opened 2026-01-13
Force TLSv1.3 on iOS

smorkolama:


AfanasievAnton (commented):
Could you please explain the rationale for having configurable TLS 1.3 on iOS? I’d like to better understand the underlying need and use cases.

AfanasievAnton:
Thanks for the feedback and much appreciate your pr :)

I see this issue was raised by your team via email channel. We will take a closer look and use the email channel to follow up on it. 

Thanks again!

AfanasievAnton:
@smorkolama Configurable minimum TLS version for WebSocket connections is now available in latest 2.13.0 release of transport

PR #504, opened 2026-03-22
[MTSDK-928] Emit Event.Logout when WS is disconnected during logoutFromAuthenticatedSession

swienribGenesys:
Add onLogoutSuccess callback to AuthHandler.logout() so that when the REST logout succeeds and the WebSocket is already closed, Event.Logout is emitted directly instead of waiting for a Shyrka LogoutEvent that will never arrive.

### DIFF
--- a/transport/GenesysCloudMessengerTransport-Dev.podspec
+++ b/transport/GenesysCloudMessengerTransport-Dev.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'GenesysCloudMessengerTransport'
-    spec.version                  = '2.12.0-rc3'
+    spec.version                  = '2.12.0'
     spec.homepage                 = 'https://github.com/MyPureCloud/genesys-messenger-transport-mobile-sdk'
     spec.source                   = { :http => '' }
     spec.authors                  = 'Genesys Cloud Services, Inc.'
--- a/transport/GenesysCloudMessengerTransport-Dev.podspec
+++ b/transport/GenesysCloudMessengerTransport-Dev.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'GenesysCloudMessengerTransport'
-    spec.version                  = '2.11.0'
+    spec.version                  = '2.12.0-rc3'
     spec.homepage                 = 'https://github.com/MyPureCloud/genesys-messenger-transport-mobile-sdk'
     spec.source                   = { :http => '' }
     spec.authors                  = 'Genesys Cloud Services, Inc.'
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
@@ -362,7 +362,7 @@ internal class MessagingClientImpl(
                     performLogout()
                 }
             },
-            onLogoutFailure = {
+            onUnauthorized = {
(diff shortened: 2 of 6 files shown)


transport/GenesysCloudMessengerTransport-Dev.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'GenesysCloudMessengerTransport'
-    spec.version                  = '2.11.0'
+    spec.version                  = '2.12.0-rc3'
```
yarin-sason: It should be 2.12.0 now (the version has been released)

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -356,9 +356,16 @@ internal class MessagingClientImpl(
     }
 
     override fun logoutFromAuthenticatedSession() {
-        authHandler.logout {
-            performLogout()
-        }
+        authHandler.logout(
+            onLogoutSuccess = {
+                if (stateMachine.isInactive()) {
```
swienribGenesys, line 361: Websocket is closed

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -786,7 +793,9 @@ internal class MessagingClientImpl(
         vault.wasAuthenticated = false
         authHandler.clear()
         eventHandler.onEvent(Event.Logout)
-        disconnect()
+        if (!stateMachine.isInactive()) {
```
swienribGenesys, line 796: Websocket is active

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -356,9 +356,16 @@ internal class MessagingClientImpl(
...
-            performLogout()
-        }
+        authHandler.logout(
+            onLogoutSuccess = {
+                if (stateMachine.isInactive()) {
+                    performLogout()
+                }
+            },
+            onLogoutFailure = {
+                performLogout()
```
yarin-sason, line 366: @swienribGenesys
Why if there was a failure we still want to pass the LogOut event to the client?

swienribGenesys: This is existing behavior, not new. The onLogoutFailure callback is only invoked when the REST call returns 401 (unauthorized), meaning the session is already expired server-side. In that case, the user is effectively already logged out, so we clean up locally and emit Event.Logout. This was the implicit flow TTL fix from 2.11.

yarin-sason: Oh I see, 
So in this point will might just think about renaming the callback, a bit confusing 😅
But whatever you pick

yarin-sason (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
yarin-sason: Suggestion - onUnauthorized

janoskernacsgenesys (approved):

PR #506, opened 2026-03-23
MTSDK-1323 Time Slot Picker: add originatingMessageId

janoskernacsgenesys:


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ButtonResponse.kt
```diff
@@ -14,4 +14,8 @@ data class ButtonResponse(
     val text: String,
     val payload: String,
     val type: String,
-)
+    val originatingMessageId: String? = null
```
swienribGenesys, line 19: Shouldn't you add originatingMessageId also to the KDoc?

swienribGenesys:
PR title has "SLot" instead of "Slot" (=

swienribGenesys (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/ButtonResponse.kt
janoskernacsgenesys: fixed

yarin-sason (approved):

PR #508, opened 2026-03-31
[MTSDK-1385] Add configurable base URL to Transport layer for UI testing

swienribGenesys:
## Description
Add optional `customEndpoint ` parameter (host:port) to the Transport layer, allowing UI test apps to route connections to a local server (e.g., WireMock) instead of the production backend. When not set, behavior is unchanged.

### Prerequisites
- Checkout the PoC branches for each platform:
  - **iOS:** https://bitbucket.org/inindca/mobile-messenger-ios-sdk/src/8c214f8ce5f1b277260c29b644338ad9501eacfc/?at=poc%2FMTSDK-1385-transport-configurable-base-url-for-ui-testing
  - **Android:** https://bitbucket.org/inindca/mobile-messenger-android-sdk/src/62daad0540a94e704f84581cdf22576e442f7899/?at=poc%2FMTSDK-1385-transport-configurable-base-url-for-ui-testing
- Configure each platform to use Transport as a **local dependency** so the changes in this PR are picked up:
  - **iOS:** https://genesys-confluence.atlassian.net/wiki/spaces/GCDMOB/pages/1015644508/Embedding+Transport+SDK+as+a+Local+Dependency+Debug+its+Kotlin+Multiplatform+Code+in+Xcode
  - **Android:** https://genesys-confluence.atlassian.net/wiki/spaces/GCDMOB/pages/1511819544/Using+Transport+as+a+Local+Dependency+Android?focusedCommentId=1613795690
### iOS
1. Open the `GCMessengerUITester` project in Xcode.
2. Run the UI test `test_customEndpoint_connectionFailsWithLocalServer ` from the test navigator.
3. Verify the test passes - you can find in the logs that it confirms Transport routed to `localhost:8080` and surfaced a connection error.
### Android
1. Open the project in Android Studio with an emulator running.
2. Run the instrumented tests in `ui-tester/src/androidTest/.../CustomEndpointTest.kt` from the test navigator.
3. Verify both tests pass:
   - `customEndpoint_connectionFailsWithLocalServer ` - confirms Transport routed to `localhost:8080`.
   - `defaultEndpoint_noOverrideWhenNotSet ` - confirms production behavior is unchanged when no override is set.
### Both Platforms
- Verify that existing unit tests in Transport (`UrlsTest`, `ConfigurationTest`, `InternalConfigurationFactoryTest`) pass with no regressions.
- Verify that the sample apps build and run normally without any custom base URL set (no breaking change).


AndrasSolyomGenesys (approved):

AndrasSolyomGenesys:
Direct transport customers might want to do similar testing of their UI implementation. Why don't we open this possibility up and have this as a documented feature?

yarin-sason:
@AndrasSolyomGenesys @swienribGenesys 
 I still don’t fully understand why we don’t mock transport and avoid adding this change only for testing purposes? If the intention is to test transport as well, then I guess it would make sense.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalConfigurationFactory.kt
```diff
@@ -28,7 +29,8 @@ object InternalConfigurationFactory {
         autoRefreshTokenWhenExpired: Boolean = true,
         encryptedVault: Boolean = false,
         sessionExpirationNoticeIntervalSeconds: Long = Configuration.DEFAULT_INTERVAL,
-        minimumWebSocketTlsVersion: TlsVersion = TlsVersion.SYSTEM_DEFAULT
+        minimumWebSocketTlsVersion: TlsVersion = TlsVersion.SYSTEM_DEFAULT,
+        customBaseUrl: String? = null
```
yarin-sason: We should have here as the default the real base url that could be overridden (instead of null)

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/Urls.kt
```diff
@@ -10,23 +10,34 @@ private const val BASE_WEBMESSAGING_PATH = "api/v2/webmessaging"
 internal class Urls(
     val domain: String,
     val deploymentId: String,
-    val application: String
+    val application: String,
+    customBaseUrl: String? = null
```
yarin-sason: in my opinion it should be nullable, we would inject the url to the URLs object,

swienribGenesys:
> @AndrasSolyomGenesys @swienribGenesys I still don’t fully understand why we don’t mock transport and avoid adding this change only for testing purposes? If the intention is to test transport as well, then I guess it would make sense.

The intention is exactly that - to test Transport as well. We want end-to-end coverage from the UI layer through the real Transport code (WebSocket connection, Shyrka message parsing, reconnection, etc.), with only the backend being simulated by WireMock. If we mock Transport itself, we'd only be testing the UI in isolation and wouldn't catch regressions in how Transport handles actual protocol messages. This approach gives us confidence in the full client stack.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalConfigurationFactory.kt
swienribGenesys: There's no single production base URL to use as a default - it's derived dynamically from domain and differs per endpoint (wss://webmessaging.$domain, https://api.$domain, https://api-cdn.$domain). So null acts as the sentinel for "use standard production URLs," and we only override when a custom URL is explicitly provided.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/Urls.kt
swienribGenesys: That's what we have - customBaseUrl is already nullable with a default of null. When null, all URLs fall back to the standard production endpoints.

yarin-sason (approved):

mobichel:
> @AndrasSolyomGenesys @swienribGenesys I still don’t fully understand why we don’t mock transport and avoid adding this change only for testing purposes? If the intention is to test transport as well, then I guess it would make sense.

I think we should be careful with making anything like this as public especially without explicit request from customer or product. Once it becomes public we need to maintain this functionality and avoid "breaking" changes.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalConfigurationFactory.kt
```diff
@@ -16,6 +16,7 @@ object InternalConfigurationFactory {
      * @param encryptedVault indicates if encrypted vault should be used.
      * @param minimumWebSocketTlsVersion the minimum TLS protocol version for WebSocket connections.
      * @param sessionExpirationNoticeIntervalSeconds how many seconds before the session expires to show the expiration notice.
+     * @param customBaseUrl optional base URL override (host:port) for routing connections to a local server (e.g., WireMock). Intended for UI testing only.
```
mobichel: `host:port` is not an URL because it's missing scheme part (e.g. https, http, etc). 
If you don't need to customize scheme part (from http to http), I would suggest adding two optional parameters instead: `host` and `port`

mobichel (changes requested):

mobichel:
> > @AndrasSolyomGenesys @swienribGenesys I still don’t fully understand why we don’t mock transport and avoid adding this change only for testing purposes? If the intention is to test transport as well, then I guess it would make sense.
> 
> The intention is exactly that - to test Transport as well. We want end-to-end coverage from the UI layer through the real Transport code (WebSocket connection, Shyrka message parsing, reconnection, etc.), with only the backend being simulated by WireMock. If we mock Transport itself, we'd only be testing the UI in isolation and wouldn't catch regressions in how Transport handles actual protocol messages. This approach gives us confidence in the full client stack.

I agree with Shahar, consider Mobile Messenger SDK with Transport SDK as single component and it's going to be tested with the external dependencies mocked.
 
According to https://genesys-confluence.atlassian.net/wiki/spaces/PureCloud/pages/325090227/Genesys+Cloud+Testing+Pyramid each service should have unit, integration and system level tests. 

If we would consider Mobile Messenger SDK and Transport SDK separately, it means we need to write integration tests  twice - for both Mobile Messenger SDK and Transport SDK and with WebSocket connection still be mocked for the last.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalConfigurationFactory.kt
swienribGenesys: **Good point about the naming, it's not technically a URL. However, I'd prefer keeping it as a single String? rather than splitting into host + port for a few reasons:**
1. WireMock (our primary use case) is addressed as host:port - callers naturally think of it as one value (e.g., "localhost:8080"), and that's exactly what gets passed via environment variables from the test runner. Splitting it would require callers to parse/split the string themselves before passing it in.
2. A single string is simpler to thread through the layers - Configuration → InternalConfigurationFactory → Urls → iOS/Android consumer code. Two parameters doubles the surface area for no functional benefit, since they're always used together and never independently.
3. sanitizeBaseUrl() already handles malformed input gracefully - if someone passes "http://localhost:8080/", the scheme and trailing slash are stripped. This is more forgiving than requiring callers to get two separate params exactly right.

I can rename it to something like customServerHost to avoid the "URL" confusion, if that helps clarify that it's not a full URL. WDYT?

mobichel: I would suggest `customEndpoint` name — most common and widely understood or `customHostPort` as more verbose.

mobichel (approved):

PR #515, opened 2026-05-06
[MTSDK-1409] Read TRANSPORT_CUSTOM_ENDPOINT env var as fallback for customEndpoint

swienribGenesys:


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/InternalConfigurationFactory.kt
```diff
@@ -49,7 +49,7 @@ object InternalConfigurationFactory {
                 ApplicationType.TRANSPORT_SDK -> "TransportSDK-${MessengerTransportSDK.sdkVersion}"
                 else -> "${applicationType.title}-$applicationVersion/TransportSDK-${MessengerTransportSDK.sdkVersion}"
             }
-        config.customEndpoint = customEndpoint
+        config.customEndpoint = customEndpoint ?: config.customEndpoint
```
AndrasSolyomGenesys: whats the point of having 2 ways to provide custom endpoint? why not just env variable?

swienribGenesys: No need, it's was leftovers from MTSDK-1385 - removed

AndrasSolyomGenesys (approved):

transport/src/androidMain/kotlin/com/genesys/cloud/messenger/transport/util/Environment.kt
```diff
@@ -0,0 +1,6 @@
+@file:JvmName("AndroidEnvironment")
+
+package com.genesys.cloud.messenger.transport.util
+
+internal actual fun getEnvironmentVariable(name: String): String? =
```
yarin-sason, line 5: @swienribGenesys 

How would this mechanism work if we want to test the messenger as part of the pipeline? 
We will want to set them as part of the pipeline? and would they be cleaned after the execution?

swienribGenesys: On iOS, `XCUIApplication.launchEnvironment` sets env vars only for the app process being launched - they're scoped to that launch and gone when the app terminates. On Android, `System.getenv() `reads the process environment which is similarly scoped. So yes, in CI the test runner sets the env var before launch, and it's automatically cleaned up when the process ends - nothing persists.

yarin-sason: sounds good thanks!

yarin-sason (approved):

abarshev (approved):

PR #527, opened 2026-06-22
MTSDK-1506 wire JourneyContext onto MessagingClient (guest + auth) 

yarin-sason:


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/JourneyContextBuilder.kt
```diff
@@ -0,0 +1,28 @@
...
+ * or when the provider yields `null` so that callers can omit the `journeyContext`
+ * field from the serialized payload.
+ *
+ * `customer.idType` is always `"cookie"`. `customerSession.type` is always `"app"`
+ * and `customerSession` itself is omitted when [JourneyContextInfo.sessionId] is `null`.
+ */
+internal fun buildJourneyContext(
+    provider: (() -> JourneyContextInfo?)?,
+): JourneyContext? =
+    provider?.invoke()?.let { info ->
```
swienribGenesys: `buildJourneyContext` invokes the consumer's `journeyContextProvider` unguarded - on the socket `onOpen` thread for guest configure, and on the caller's thread in `authorize`. A throwing provider would propagate into those paths. Consider wrapping the invocation (catch → log → treat as no journey context). Consistent with how other consumer callbacks are handled today, so optional, but cheap insurance at a trust boundary.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/JourneyContextBuilder.kt
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Invokes [provider] (if non-null) and maps its [JourneyContextInfo] result to the
+ * internal [JourneyContext] wire DTO. Returns `null` when no provider is supplied
+ * or when the provider yields `null` so that callers can omit the `journeyContext`
+ * field from the serialized payload.
+ *
+ * `customer.idType` is always `"cookie"`. `customerSession.type` is always `"app"`
+ * and `customerSession` itself is omitted when [JourneyContextInfo.sessionId] is `null`.
+ */
```
swienribGenesys, line 21: The ticket asks for `buildJourneyContext()` inside `MessagingClientImpl`, but it's a separate util file instead. That's actually a good call since both `MessagingClientImpl` and `AuthHandlerImpl` use it - just flagging the difference so it's not mistaken for a missed requirement.

yarin-sason: Thanks, yes make sense the reason that I moved it outside is that we need this logic in both messaging client and in the authHandler

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/JourneyContextBuilder.kt
yarin-sason: Addressed

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
```diff
@@ -100,6 +100,17 @@ interface MessagingClient {
...
 
+    /**
+     * Optional provider invoked at session-configure time to attach a [JourneyContextInfo]
+     * (customer cookie id and optional journey session id) to guest and authenticated
+     * configure requests. When `null` (or when it returns `null`), no journey context is sent.
+     *
+     *
+     * Note: the provider is invoked synchronously on the thread that builds the configure
+     * or `AuthJwtRequest` payload; it must be cheap and non-blocking.
+     */
```
swienribGenesys, line 114: Could you add a line here noting that if the provider throws, it's caught and the journey context is just omitted? Good for consumers to know a faulty provider won't break the session.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/JourneyContextBuilder.kt
```diff
@@ -0,0 +1,39 @@
...
+internal fun buildJourneyContext(
+    provider: (() -> JourneyContextInfo?)?,
+    log: Log? = null,
+): JourneyContext? {
+    if (provider == null) return null
+    val info = try {
+        provider()
+    } catch (t: Throwable) {
```
swienribGenesys: Minor: `catch (Throwable)` also catches things like OutOfMemoryError. Since this call is synchronous (no coroutine cancellation to worry about), `catch (Exception)` would be slightly more conventional. Non-blocking.

swienribGenesys (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
yarin-sason: Added

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/util/JourneyContextBuilder.kt
yarin-sason: Addressed

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -781,11 +781,6 @@ internal class MessagingClientImpl(
                 token = token,
                 deploymentId = configuration.deploymentId,
                 startNew = startNew,
-                journeyContext =
-                    JourneyContext(
-                        JourneyCustomer(token, "cookie"),
-                        JourneyCustomerSession("", "web")
-                    ),
```
janoskernacsgenesys, line 788: `buildJourneyContext` is called for the guest configure path but not here. The journeyContext field was also removed outright from ConfigureAuthenticatedSessionRequest. This means journey context is only sent during fetchAuthJwt, not during configureAuthenticatedSession. In a reconnect scenario the JWT may already be cached, so fetchAuthJwt is never called again — the server only gets journey context on the very first auth, not on subsequent reconnects. If this is intentional (the server ties the journey context to the JWT), it should be documented with a comment. If it's a gap, buildJourneyContext(journeyContextProvider, log) needs to be passed here too (and the field re-added to ConfigureAuthenticatedSessionRequest).

yarin-sason: Yes, it was intentional (Following request, we got the form product). We requested to pass the journey context only when authorizing, and not each session that has been configured (for auth session for guest that is the case). The stitching is reliant on the same cookie that is not supposed to be changed (unless the user asked it explicitly).

janoskernacsgenesys (approved):

PR #528, opened 2026-06-23
MTSDK-1511 Fix crash on attempt to disconnect in case of wrong state

janoskernacsgenesys:
Fixed iOS crash in MessagingClientImpl.SocketListener.onClosing.

And also made the disconnect crash-safe by just being a no-op in case of being in wrong states.

## Analysis on the affected states
connect() is the only way back from all three states:
```
Idle   ──connect()──→ Connecting → Connected → Configured
Closed ──connect()──→ Connecting → ...
Error  ──connect()──→ Connecting → ...
```

There is no "resume" path. disconnect() in these states has no session to close and no path back to one — so the no-op is the correct and complete behavior.

### DIFF
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
@@ -274,8 +274,10 @@ interface MessagingClient {
 
     /**
      * Close the WebSocket connection to the Web Messaging service.
-     * If the client is already disconnected (Closed, Idle, or Error state), this is a no-op.
+     *
+     * @throws IllegalStateException If the current state of the MessagingClient is not compatible with the requested action.
      */
+    @Throws(IllegalStateException::class)
     fun disconnect()
 
     /**
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
@@ -274,10 +274,8 @@ interface MessagingClient {
 
     /**
      * Close the WebSocket connection to the Web Messaging service.
-     *
-     * @throws IllegalStateException If the current state of the MessagingClient is not compatible with the requested action.
+     * If the client is already disconnected (Closed, Idle, or Error state), this is a no-op.
      */
-    @Throws(IllegalStateException::class)
     fun disconnect()
 
     /**
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
@@ -218,7 +218,6 @@ internal class MessagingClientImpl(
         closeAllConnectionsForTheSession()
     }
 
-    @Throws(IllegalStateException::class)
     override fun disconnect() {
         log.i { LogMessages.DISCONNECT }
         if (!stateMachine.canDisconnect()) {
--- a/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
+++ b/transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
@@ -218,12 +218,9 @@ internal class MessagingClientImpl(
         closeAllConnectionsForTheSession()
     }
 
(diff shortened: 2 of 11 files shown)


transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClientImpl.kt
```diff
@@ -934,6 +938,10 @@ internal class MessagingClientImpl(
             reason: String
         ) {
             log.i { LogMessages.onClosing(code, reason) }
+            if (!stateMachine.canDisconnect()) {
```
AfanasievAnton: Before making this change - lets validate the side effects of never reaching the stateMachine.onClosing(code, reason)

janoskernacsgenesys: Here we don't need the middle step with state Closing. 
So with this fix when we skip the → Closing step for already-terminal states, it's not skipping any work — it's skipping a signaling step that has no meaning when the connection is already dead.


By the time the state is Closed, Idle, or Error, onClosed has already happened (or never needed to happen):

- Closed — got here via onClosed firing → stateMachine.onClosed() + cleanUp(). Already done.
- Error — got here via onFailure → deactivateAndCancelWebSocket() (socket already cancelled) + transitionToStateError(). Socket is already gone, resources already cleared.
- Idle — initial state. No socket was ever opened, so onClosed was never relevant.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
```diff
@@ -274,10 +274,8 @@ interface MessagingClient {
 
     /**
      * Close the WebSocket connection to the Web Messaging service.
-     *
-     * @throws IllegalStateException If the current state of the MessagingClient is not compatible with the requested action.
+     * If the client is already disconnected (Closed, Idle, or Error state), this is a no-op.
      */
-    @Throws(IllegalStateException::class)
```
AndrasSolyomGenesys, line 280: What is our definition of a breaking change?
Removing throws will not result in a build error, only a warning: `No calls to throwing functions occur within 'try' expression`
However when very closely following semantic versioning rules it can be considered a breaking change as the behavior of the function changes.

Can we just fix the function to correctly throw an Error instead of crashing, instead of removing the throw?

No-op behavior in general are not good design I think, there is no feedback to consumer why something did not happen.

janoskernacsgenesys: No-op behavior in this case seems to be correct because the attempts of disconnect should not end in error  in a state where the connection is already closed

GolanSG:
Now that both callers (`disconnect()` and `SocketListener.onClosing`) guard with `canDisconnect()` before reaching `stateMachine.onClosing()` — the check(`currentState.canDisconnect()`) inside `StateMachineImpl.onClosing()` is effectively dead code.

Not a blocker, but worth considering: keep it as a defensive safety net (and maybe add a short comment saying so), or remove it to avoid confusing future readers who see a throw that can never be reached in practice.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessagingClient.kt
janoskernacsgenesys: > No-op behavior in this case seems to be correct because the attempts of disconnect should not end in error in a state where the connection is already closed

At least on Android/Kotlin/JVM side. 
iOS side may be different

AndrasSolyomGenesys: It is breaking change for obj-c host apps, they will fail build with error if throwing is removed
in this case solution could be: no-on, but keep throwing, deprecating the throw and remove in future

GolanSG: I'm with Andras's opinion that removing the throw results in potentially bad user experience, since the user won't have indication that they're doing something wrong (trying to disconnect when the session was already disconnected). 
@janoskernacsgenesys was this change needed in order for this crash to be fixed?

janoskernacsgenesys: @GolanSG @AndrasSolyomGenesys Please check the updated solution with keeping the throw

AndrasSolyomGenesys (approved):

GolanSG (approved):

PR #533, opened 2026-08-25
MTSDK-1572 host JourneyContext on MessengerTransportSDK and use journ…

yarin-sason:
…ey customer for push registration

### DIFF
--- a/transport/src/androidUnitTest/kotlin/transport/core/MessengerTransportSDKTest.kt
+++ b/transport/src/androidUnitTest/kotlin/transport/core/MessengerTransportSDKTest.kt
@@ -3,6 +3,7 @@ package transport.core
 import android.content.Context
 import assertk.assertThat
 import assertk.assertions.isInstanceOf
+import com.genesys.cloud.messenger.transport.core.JourneyContextInfo
 import com.genesys.cloud.messenger.transport.core.MessagingClient
 import com.genesys.cloud.messenger.transport.core.MessengerTransportSDK
 import com.genesys.cloud.messenger.transport.push.PushService
@@ -76,5 +77,17 @@ class MessengerTransportSDKTest {
         val result = subject.createPushService()
 
         assertThat(result).isInstanceOf(PushService::class.java)
+    }
+
+    @Test
+    fun `when MessengerTransportSDK() is created with provided vault and journeyContextProvider`() {
+        val fakeVault = FakeVault(TestValues.vaultKeys)
+        val givenProvider: () -> JourneyContextInfo? = { null }
+
+        val subject = MessengerTransportSDK(TestValues.configuration, fakeVault, givenProvider)
+
+        assertThat(subject.vault).isInstanceOf(FakeVault::class.java)
+        assertThat(subject.createPushService()).isInstanceOf(PushService::class.java)
+        assertThat(subject.createMessagingClient()).isInstanceOf(MessagingClient::class.java)
     }
 }
--- a/transport/src/androidUnitTest/kotlin/transport/core/messagingclient/BaseMessagingClientTest.kt
+++ b/transport/src/androidUnitTest/kotlin/transport/core/messagingclient/BaseMessagingClientTest.kt
@@ -6,6 +6,7 @@ import com.genesys.cloud.messenger.transport.core.ButtonResponse
 import com.genesys.cloud.messenger.transport.core.CustomAttributesStoreImpl
 import com.genesys.cloud.messenger.transport.core.Empty
 import com.genesys.cloud.messenger.transport.core.HistoryHandler
+import com.genesys.cloud.messenger.transport.core.JourneyContextInfo
 import com.genesys.cloud.messenger.transport.core.JwtHandler
 import com.genesys.cloud.messenger.transport.core.Message
 import com.genesys.cloud.messenger.transport.core.MessageStore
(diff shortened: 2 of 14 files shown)


AfanasievAnton (approved):

janoskernacsgenesys (approved):

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessengerTransportSDK.kt
```diff
@@ -24,11 +24,15 @@ import kotlinx.coroutines.withContext
  * The entry point to the services provided by the transport SDK.
  *
  * @param vault the storage mechanism for managing session-related data.
+ * @param journeyContextProvider optional provider invoked at session-configure and push-registration
+ *        time to attach a device-scoped customer cookie id. `null` (or returning `null`) omits
+ *        journey context and keeps push identity as the session token.
```
swienribGenesys, line 29: nit: the doc we deleted from `MessagingClient.journeyContextProvider` also said the provider should be cheap and non-blocking, and that if it throws we log it and carry on without journey context. Worth keeping those two lines here since this is now the only place a consumer would see them. Separately, `JourneyContextInfo`'s own KDoc still says it's only for session-configure - it sets the push registration identity now too.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/push/PushServiceImpl.kt
```diff
@@ -159,7 +162,7 @@ internal class PushServiceImpl(
         pushProvider: PushProvider,
     ): PushConfig {
         return PushConfig(
-            token = vault.token,
+            token = buildJourneyContext(journeyContextProvider, log)?.customer?.id ?: vault.token,
```
swienribGenesys: nit: `buildJourneyContext` builds the JSON object we send on configure/auth, so here we wrap the cookie id into `JourneyContext(customer = JourneyCustomer(id = ...))` and immediately unwrap it again with `?.customer?.id`. What we actually want to reuse is the try/catch + warning log. Could we split that into a small `journeyContextInfoOrNull(provider, log)`, have `buildJourneyContext` call it, and read `?.customerCookieId` here? Reads clearer and keeps push independent of the messaging DTO.

transport/src/androidUnitTest/kotlin/transport/core/MessengerTransportSDKTest.kt
```diff
@@ -77,4 +78,16 @@ class MessengerTransportSDKTest {
...
+    @Test
+    fun `when MessengerTransportSDK() is created with provided vault and journeyContextProvider`() {
+        val fakeVault = FakeVault(TestValues.vaultKeys)
+        val givenProvider: () -> JourneyContextInfo? = { null }
+
+        val subject = MessengerTransportSDK(TestValues.configuration, fakeVault, givenProvider)
+
+        assertThat(subject.vault).isInstanceOf(FakeVault::class.java)
+        assertThat(subject.createPushService()).isInstanceOf(PushService::class.java)
+        assertThat(subject.createMessagingClient()).isInstanceOf(MessagingClient::class.java)
```
swienribGenesys, line 91: I deleted all four lines that pass the provider down (`MessagingClientImpl` 99 and 111, `MessengerTransportSDK` 139 and 167) and the whole suite still passed: 844 tests, 0 failures, same as before. Nothing covers the wiring - `BaseMessagingClientTest` always injects `mockAuthHandler`/`mockPushService` so lines 99 and 111 never run, and this test only checks the return types, so `givenProvider` could be unused and it would still be green. Could we add one test that builds `MessagingClientImpl` without injecting `authHandler`/`pushService` and asserts the cookie reaches the api? These four lines are exactly what a future merge conflict drops silently.

transport/src/androidUnitTest/kotlin/transport/core/messagingclient/BaseMessagingClientTest.kt
```diff
@@ -270,6 +271,32 @@ open class BaseMessagingClientTest {
             it.stateChangedListener = mockStateChangedListener
         }
 
+    internal fun buildSubject(journeyContextProvider: (() -> JourneyContextInfo?)? = null): MessagingClientImpl =
```
swienribGenesys, line 251: nit: this is the same 22 lines as `subject` above (line 249), apart from `journeyContextProvider`. Could `subject` just be `buildSubject()`? Otherwise any future constructor change has to be made in both places, and ~15 test files inherit this.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/push/PushServiceImpl.kt
swienribGenesys: Separate point on this same line: `customerCookieId` is a non-null `String`, so `?: vault.token` only kicks in when the whole `JourneyContextInfo` is null. A provider returning `JourneyContextInfo("", null)` registers the device with an empty token, and the comparator then reads that as `Diff.TOKEN` on every sync. If you take the `journeyContextInfoOrNull` split above, `?.customerCookieId?.takeIf { it.isNotBlank() }` covers it in the same edit.

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/core/MessengerTransportSDK.kt
yarin-sason: Done

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/push/PushServiceImpl.kt
yarin-sason: Done

transport/src/androidUnitTest/kotlin/transport/core/MessengerTransportSDKTest.kt
yarin-sason: Done

transport/src/androidUnitTest/kotlin/transport/core/messagingclient/BaseMessagingClientTest.kt
yarin-sason: Done

transport/src/commonMain/kotlin/com/genesys/cloud/messenger/transport/push/PushServiceImpl.kt
yarin-sason: Done

swienribGenesys (approved):

PR #534, opened 2026-08-27
MTSDK-1565 Unsynchronized EncryptedVault access causes crashes

janoskernacsgenesys:


transport/src/androidUnitTest/kotlin/transport/core/InternalVaultTest.kt
```diff
@@ -21,12 +24,173 @@ import org.junit.Before
...
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicBoolean
 import javax.crypto.Cipher
 import javax.crypto.KeyGenerator
 import javax.crypto.SecretKey
 import javax.crypto.spec.GCMParameterSpec
 
 class InternalVaultTest {
+    @Test
+    fun `when store() runs concurrently on instances sharing storage`() {
```
swienribGenesys, line 39: Could we add the concurrency stress test required by the AC? The current tests verify the lock with mocked dependencies and only a few operations, but they do not repeatedly exercise the real Android Keystore and SharedPreferences path that caused the native crash.

yarin-sason (approved):

SzaszaGenesys (approved):

transport/src/androidUnitTest/kotlin/transport/core/InternalVaultTest.kt
janoskernacsgenesys: It is not possible to do it in the scope of this ticket. The validation should continue in the related GMMS ticket

janoskernacsgenesys: So I have moved target branch of the pr and I have created new RC for Android. So mobile part can proceed

PR #535, opened 2026-08-31
[Do not merge] wmm-965 pc for new-auth-path implementation

yarin-sason:


jeltehomminga:
Hope this one doesn't stay a PoC — it fixes a live failure.

Replaying the exchange endpoint (eu-central) with one fixed ID token, changing only `oauth.nonce`:

| `oauth.nonce` | result |
| --- | --- |
| absent | 200 |
| `null` | 200 |
| exact match with the claim | 200 |
| `""` | 401 `unauthorized` |
| anything else | 401 |

Omitted, null, or exactly right. Empty string never works.

Host apps can't avoid sending it. `nonce` is a non-null `String` in `authorizeImplicit`, `AuthHandler`
and `fetchAuthJwt`, and again in the Messenger SDK's `setImplicitAuthenticationInfo` and
`reauthorizeImplicitFlow`. If the IdP issued a token with no `nonce` claim, there's nothing left to
pass but `""`. The wire model is already fine — `OAuth.nonce` is `String? = null` and `encodeDefaults`
is off, so null gets dropped from the body. It's only the signatures.

This isn't a rare case either. Auth0 puts `nonce` in the token only when the authorize request carried
one, and a refresh grant has no authorize request. So authenticated messaging works right after
sign-in and dies at the first silent renewal, on any Auth0 tenant. Same root cause as
[this thread](https://community.genesys.com/discussion/android-mobile-messenger-401-authfailed-during-oauth-code-exchange-pkce),
where the fix was setting the nonce to null.

The docs currently point the other way, for what it's worth: `WebDeploymentsOAuthRequestParameters`
says `nonce` "should be empty" when it wasn't used, and that's the 401.

Two asks: ship this, and mirror it into the Messenger SDK's `setImplicitAuthenticationInfo` and
`reauthorizeImplicitFlow` on both platforms. That SDK is closed source and it's what host apps
actually call, so merging this alone won't unblock anyone using it.

